Issue #6, opened 2025-11-14
Create Research Supervisor


sugat009:
Create Research Supervisor agent using LangChain under which there would be two agents:
1. Documentation Agent
2. Context Agent

Design [here](https://github.com/Hareet/cht-agent/blob/main/design/CHT%20Hierarchical%20Multi-Agent%20System%20PoC.md).

PR #7, opened 2025-11-19
feat(#6): add research supervisor and its reports documentation agent and context agent

sugat009:
# Description

Creates research supervisor and it's reports the documentation agent(output mocked for now) and the context agent. Tested this for a dummy ticket `tickets/simple-example.md` and for a real `cht-core` [ticket](https://github.com/medic/cht-core/issues/10139) `tickets/10139.md`. _The output of the research might seem a bit basic at the moment due to lack of prior context and solved tickets._

medic/cht-agent#6

# Code review checklist

- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.
 
# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

### DIFF
--- a/src/supervisors/research-supervisor.ts
+++ b/src/supervisors/research-supervisor.ts
@@ -18,7 +18,6 @@ import {
 } from '../types';
 import { DocumentationSearchAgent } from '../agents/documentation-search-agent';
 import { ContextAnalysisAgent } from '../agents/context-analysis-agent';
-import { enrichIssueTemplate } from '../utils/domain-inference';
 
 // Define the state annotation for type safety
 const ResearchStateAnnotation = Annotation.Root({
@@ -447,9 +446,13 @@ Format your response as a structured plan that will guide the development team.`
 
     const hours = baseHours[complexity] * (phaseCount / 4);
 
-    if (hours < 8) return `${hours} hours`;
-    if (hours < 40) return `${Math.round(hours / 8)} days`;
-    return `${Math.round(hours / 40)} weeks`;
+    if (hours < 8) return `${hours} hour${hours === 1 ? '' : 's'}`;
+    if (hours < 40) {
+      const days = Math.round(hours / 8);
+      return `${days} day${days === 1 ? '' : 's'}`;
+    }
+    const weeks = Math.round(hours / 40);
+    return `${weeks} week${weeks === 1 ? '' : 's'}`;
   }
 
   /**
@@ -460,30 +463,19 @@ Format your response as a structured plan that will guide the development team.`
     console.log('RESEARCH SUPERVISOR - Starting Research Phase');
     console.log('========================================');
     console.log(`Issue: ${issue.issue.title}`);
-
-    // Enrich issue with domain/components if not specified
-    let enrichedIssue = issue;
-    if (
-      !issue.issue.technical_context.domain ||
-      issue.issue.technical_context.components.length === 0
-    ) {
-      console.log('Domain or components not specified - inferring from description...\n');
-      enrichedIssue = await enrichIssueTemplate(issue, this.plannerModel.modelName);
-    }
-
-    console.log(`Domain: ${enrichedIssue.issue.technical_context.domain}`);
-    console.log(`Components: ${enrichedIssue.issue.technical_context.components.join(', ')}`);
+    console.log(`Domain: ${issue.issue.technical_context.domain}`);
+    console.log(`Components: ${issue.issue.technical_context.components.join(', ') || 'None specified'}`);
     console.log('========================================\n');
 
     const initialState: typeof ResearchStateAnnotation.State = {
       messages: [
         {
           role: 'user',
-          content: `Research issue: ${enrichedIssue.issue.title}`,
+          content: `Research issue: ${issue.issue.title}`,
           timestamp: new Date().toISOString(),
         },
       ],
-      issue: enrichedIssue,
+      issue: issue,
       researchFindings: undefined,
       contextAnalysis: undefined,
       orchestrationPlan: undefined,
@@ -491,7 +483,24 @@ Format your response as a structured plan that will guide the development team.`
       errors: [],
     };
 
-    const result = await this.graph.invoke(initialState);
+    let result: typeof ResearchStateAnnotation.State;
+    try {
+      result = await this.graph.invoke(initialState);
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : 'Unknown error during graph execution';
+      console.error('\n========================================');
+      console.error('RESEARCH SUPERVISOR - Error during execution');
+      console.error('========================================');
+      console.error(`Error: ${errorMessage}`);
+      console.error('========================================\n');
+
+      // Return state with error recorded
+      return {
+        ...initialState,
+        currentPhase: 'error',
+        errors: [errorMessage],
+      };
+    }
 
     console.log('\n========================================');
     console.log('RESEARCH SUPERVISOR - Research Phase Complete');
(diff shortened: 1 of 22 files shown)


src/utils/ticket-parser.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  // Build IssueTemplate
+  const issueTemplate: IssueTemplate = {
+    issue: {
+      title: metadata.title,
+      type: metadata.type as 'feature' | 'bug' | 'enhancement',
+      priority: metadata.priority as 'high' | 'medium' | 'low',
+      description: descriptionSection || markdown.trim() || '',
+      technical_context: {
+        domain: metadata.domain ? validateDomain(metadata.domain) : undefined,
```
Hareet: I like what you are doing for `validateDomain`  here and back to line 72-80.

Should type and priority above have the same validations? If priority comes as `urgent`, does that sneak through? 


```typescript
 // Validate required metadata
  if (!metadata.title) {
    throw new Error('Ticket must have a "title" in frontmatter');
  }

  if (!metadata.type) {
    throw new Error('Ticket must have a "type" in frontmatter (feature|bug|enhancement)');
  }

  if (!metadata.priority) {
    throw new Error('Ticket must have a "priority" in frontmatter (high|medium|low)');
  }
``` 

```typescript
**
 * Validate that domain is a valid CHTDomain
 */
function validateDomain(domain: string): CHTDomain {
  const validDomains: CHTDomain[] = [
    'authentication',
    'contacts',
    'forms-and-reports',
    'tasks-and-targets',
    'messaging',
    'data-sync',
    'configuration',
  ];

  if (validDomains.includes(domain as CHTDomain)) {
    return domain as CHTDomain;
  }

  throw new Error(`Invalid domain: "${domain}". Must be one of: ${validDomains.join(', ')}`);
}
```

I think `metadata.title` is fine to leave alone

src/utils/domain-inference.ts
```diff
@@ -0,0 +1,176 @@
...
+  const response = await model.invoke(prompt);
+  const content = response.content.toString();
+
+  // Extract JSON from response
+  const jsonMatch = content.match(/\{[\s\S]*\}/);
+  if (!jsonMatch) {
+    throw new Error('LLM did not return valid JSON response');
+  }
+
+  const result = JSON.parse(jsonMatch[0]);
```
Hareet: Claude flagged this and it's reasoning makes sense to me in the context of we have no idea what the llm will send back to us.
```
Issue: JSON.parse can throw a SyntaxError even after regex validation. The regex \{[\s\S]*\} matches any text between braces, including malformed JSON like {broken: no-quotes}.

Impact: Application crash on malformed LLM response
```

tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "resolveJsonModule": true,
+    "moduleResolution": "node",
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
```
Hareet: nit: noImplicitAny, strictNullChecks, strictFunctionTypes are already included in [strict: true](https://www.typescriptlang.org/tsconfig/#noImplicitAny)

src/supervisors/research-supervisor.ts
```diff
@@ -0,0 +1,505 @@
...
+      low: 4,
+      medium: 16,
+      high: 40,
+    };
+
+    const hours = baseHours[complexity] * (phaseCount / 4);
+
+    if (hours < 8) return `${hours} hours`;
+    if (hours < 40) return `${Math.round(hours / 8)} days`;
+    return `${Math.round(hours / 40)} weeks`;
```
Hareet: nit: if it equals 40, we return `1 weeks`. That's also what the test wants, but grammatically should read `1 week`.

sugat009:
### LangGraph

> How are we currently feeling about LangGraph? Is it overkill for 4-5 agents with 2-3 supervisors? Should we wrap Agent SDK calls directly and not worry about node state. Do you feel comfortable enough on this path?

LangGraph is justified because of the feedback loops. Without cycles I'd agree it's overkill, but human rejection and QA failures need to route back. That's what graphs are for. I'm comfortable continuing.

---

### MCP / Kapa.ai Workflow

> I'm imagining we'll need a few llm calls here. Potential workflow: our code calls MCP, retrieves list of sources or results to ask_question. Our code makes an LLM call along with agent-memory, and then we tell it to concise it down.

Use `ask_question`. Kapa already did the heavy lifting with RAG + user context. We just need one LLM call to contextualize their answer with our agent-memory and structure the output.

---

### LLM Models

> We want to be able to use anything. Is that a plan to refactor later, or should we pull it out into a module now?

Agree we should do it now, not later. I've already built this as part of the dev supervisor work, an LLM abstraction with interface + factory pattern. Claude is fully implemented, OpenAI/Gemini are stubbed for when we need them. Will be in the next PR for review.

---

### LLM API calls vs Agent SDKs

> API calls are a bit lower level and might be missing a ton of the system prompt that are injected with Claude Code that make it a viable coding assistant. How do we imagine handling that without getting locked into a vendor llm, and without re-inventing a functional assistant while ensuring our assistant isn't worse performing than provided agent.

I think we just build it out piece by piece. The thing is, we're not trying to build a general purpose coding assistant. Our agents do narrow, CHT specific tasks. So we can write focused prompts that leverage context from agent-memory that generic assistants wouldn't have anyway. Plus we're mostly dealing with structured outputs, not open ended "write me a feature" stuff. And with human validation checkpoints, we can catch what doesn't work and iterate. It won't be perfect out of the gate, but we'll refine as we go.

---

### TODO and Planning features

> Do we know if LangGraph handles this internally? Those are some of the core features that make Claude Code multi agents work better. Sharing todo, planning, and verification lists. Does that occur with raw api calls? I see it at the end of the Supervisor phase, but should it be added for some of the other api calls?

LangGraph handles state sharing between agents. When one finishes, the next sees its output. But it doesn't do the todo/planning stuff that Claude Code does internally. Right now our supervisors produce plans, but individual agents just execute. For simpler agents like doc search, that's fine. For code generation, we might benefit from having the agent explicitly plan before generating. This could help with reasoning and make it easier to debug when things go wrong. Worth considering for the coding and test agents.

---

### Domain Inference

> This seems highly important and I'm worried about the difficulty of getting this correct, or this value guiding the agents down a wild path. What's your opinion on leaving this out of scope for now until we manually have mapped 50+ tickets? And then we can test to see how accurate our domain inference would be, or do you see that happening alongside immediately at release?

Yes, I think it would be a good idea to have the domain as a required field for a good few tickets until we solve a few of them. Once we have context built up we can then have it as optional.

---

### Human-in-the-loop

> From the research-supervisor.ts, we should present the plan and have us read over it, modify the plan file and then agree to continue.

Yeah, that is a good idea. However, we should just ask the user to use their favorite text editor to edit the file in case they want to edit it, and not implement an editor in the agent interface.

---

### Parsing Layer

> Our parsing layer using custom parsers looks like it could get out of hand. Do you think we should go with js-yaml? ticket-parser.ts and context-loader.ts

Yup, we should use it as it is already tried and tested.

PR #85, opened 2026-04-06
fix(#83): default MCP_SERVER_URL to dev Kapa AI endpoint

imtushar01:
# Description
Default MCP_SERVER_URL to https://mcp-docs.dev.medicmobile.org/mcp when not explicitly set.
Added console.log to display the MCP server URL being used when the documentation search agent makes a call.

medic/cht-agent#83

# Code review checklist
- [x] Readable: Concise, well named, follows the style guide, documented if necessary.
- [ ] Documented: Configuration and user documentation on cht-docs
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# License
The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

### DIFF
--- a/src/agents/documentation-search-agent.ts
+++ b/src/agents/documentation-search-agent.ts
@@ -15,11 +15,13 @@ import {
 
 export class DocumentationSearchAgent {
   private useMockMCP: boolean;
+  private mcpServerUrl: string;
 
-  constructor(options: { modelName?: string; useMockMCP?: boolean } = {}) {
+  constructor(options: { modelName?: string; useMockMCP?: boolean; mcpServerUrl?: string } = {}) {
     // Model will be used when MCP integration is complete
     // For now, we use mocked responses
     this.useMockMCP = options.useMockMCP !== false; // Default to true for POC
+    this.mcpServerUrl = options.mcpServerUrl ?? process.env.MCP_SERVER_URL ?? 'https://mcp-docs.dev.medicmobile.org/mcp';
   }
 
   /**
@@ -70,6 +72,8 @@ export class DocumentationSearchAgent {
     if (this.useMockMCP) {
       return this.mockKapaAIResponse(query, domain);
     }
+
+    console.log(`[Documentation Search Agent] Calling MCP server: ${this.mcpServerUrl}`);
 
     // TODO: Actual MCP implementation when server is ready
     // const mcpCall: MCPToolCall = {
--- a/src/cli/research.ts
+++ b/src/cli/research.ts
@@ -73,7 +73,7 @@ async function main() {
     console.log('🤖 Initializing Research Supervisor...\n');
     const supervisor = new ResearchSupervisor({
       modelName: 'claude-sonnet-4-20250514',
-      useMockMCP: true, // Using mocked MCP for now
+      useMockMCP: false, // Use real MCP by default 
     });
 
     // Display issue details


src/cli/research.ts
```diff
@@ -73,7 +73,7 @@ async function main() {
     console.log('🤖 Initializing Research Supervisor...\n');
     const supervisor = new ResearchSupervisor({
       modelName: 'claude-sonnet-4-20250514',
-      useMockMCP: true, // Using mocked MCP for now
+      useMockMCP: false, // Use real MCP by default
```
sugat009, line 76: **issue (blocking):** This will crash because `callKapaAI()` throws when `useMockMCP` is `false`. The real MCP call needs to be wired up before flipping this flag.

src/agents/documentation-search-agent.ts
```diff
@@ -15,11 +15,13 @@ import {
...
 export class DocumentationSearchAgent {
   private useMockMCP: boolean;
+  private mcpServerUrl: string;
 
-  constructor(options: { modelName?: string; useMockMCP?: boolean } = {}) {
+  constructor(options: { modelName?: string; useMockMCP?: boolean; mcpServerUrl?: string } = {}) {
     // Model will be used when MCP integration is complete
     // For now, we use mocked responses
     this.useMockMCP = options.useMockMCP !== false; // Default to true for POC
+    this.mcpServerUrl = options.mcpServerUrl ?? process.env.MCP_SERVER_URL ?? 'https://mcp-docs.dev.medicmobile.org/mcp';
```
sugat009, line 24: **suggestion (non-blocking):** Consider reading `process.env.MCP_SERVER_URL` in `research.ts` (where `dotenv` is loaded) and passing it via constructor options instead of reading it here. Keeps config resolution in one place.

sugat009 (commented):
Thanks for picking this up @imtushar01! The ticket may have been misinterpreted slightly. Just defaulting the URL is not enough to get real Kapa AI results because the MCP integration in `callKapaAI()` is still a TODO stub that throws `'MCP integration not yet implemented'`. So setting `useMockMCP: false` in `research.ts` will crash `npm run research`.

While working on the code generation layer (#86), we needed higher quality output from the research layer, so a full `MCPClient` was implemented there (`src/mcp/client.ts`) with real JSON-RPC calls, timeout handling, error handling, and response parsing. Since the research agent can be operated independently, I think the scope of this ticket can be extended to pull in those components so the research agent works end to end on `main`. If that feels like too much scope, the minimum requirement would be to wire up an actual call to the MCP server in `callKapaAI()` so that flipping `useMockMCP: false` doesn't crash.

Specifically from #86:
- `src/mcp/client.ts` and `src/mcp/index.ts` (the `MCPClient` class)
- The related types in `src/types/index.ts` (`MCPClientConfig`, `MCPSearchDocsParams`, etc.)
- The updated `DocumentationSearchAgent` constructor and `callKapaAI` method that uses `MCPClient` instead of the stub

The existing tests will also need updates:
- The `'should throw error when MCP is disabled and not implemented'` test asserts the current throw behavior and would need to be replaced with a test that verifies a real (or mocked HTTP) call is made
- New tests for the MCP client call (success, timeout, error responses, malformed responses)
- The `'cached' source` test may need updating depending on how the response source is reported after wiring up real calls

This would fully satisfy the acceptance criteria in #83 and reduce merge conflicts when #86 lands later.

src/agents/documentation-search-agent.ts
```diff
@@ -15,11 +15,13 @@ import {
 
 export class DocumentationSearchAgent {
   private useMockMCP: boolean;
+  private mcpServerUrl: string;
 
-  constructor(options: { modelName?: string; useMockMCP?: boolean } = {}) {
+  constructor(options: { modelName?: string; useMockMCP?: boolean; mcpServerUrl?: string } = {}) {
     // Model will be used when MCP integration is complete
     // For now, we use mocked responses
     this.useMockMCP = options.useMockMCP !== false; // Default to true for POC
```
sugat009, line 23: **suggestion (non-blocking):** `options.useMockMCP === true` would be clearer. Currently `undefined` silently becomes `true`, which means callers that don't set it get mocked responses without realizing.

src/agents/documentation-search-agent.ts
```diff
@@ -71,6 +73,8 @@ export class DocumentationSearchAgent {
       return this.mockKapaAIResponse(query, domain);
     }
 
+    console.log(`[Documentation Search Agent] Calling MCP server: ${this.mcpServerUrl}`);
```
sugat009, line 76: **issue:** `mcpServerUrl` is logged but never used in a request. The method throws right below. This needs a real MCP client call using the URL.

src/agents/documentation-search-agent.ts
```diff
@@ -286,4 +290,4 @@ export class DocumentationSearchAgent {
 
     return relatedDomains;
   }
-}
+}
```
sugat009, line 293: **nitpick (non-blocking):** Trailing newline was removed.

imtushar01:
Thanks for the detailed review @sugat009! You're right I had misread the ticket scope. I'll extend this PR to wire up a real MCP call in `callKapaAI()` rather than just defaulting the URL.

I'll pull in the relevant pieces from #86:
- `src/mcp/client.ts` and `src/mcp/index.ts` (the `MCPClient` class)
- The related types (`MCPClientConfig`, `MCPSearchDocsParams`, etc.)
- Updated `DocumentationSearchAgent` constructor and `callKapaAI` to use `MCPClient`

For tests, I'll:
- Replace the `'should throw error when MCP is disabled and not implemented'` test with one that verifies a real (or mocked HTTP) call is made
- Add new tests for success, timeout, error responses, and malformed responses
- Update the `'cached' source` test as needed

Will push an update shortly.

sugat009:
Cool, thanks @imtushar01 !

Issue #67, opened 2026-03-03
Implement Test Orchestration Layer


sugat009:
## Summary

Implement the Test Orchestration Layer as a thin wrapper around npm scripts, based on discovery findings from #18.

## Background

After code and tests are generated, this layer runs the tests and reports results. CHT already has test infrastructure; we orchestrate it.

### 1. Thin Wrapper Around npm Scripts

```typescript
async function runTests(targetProject: string, scope: string): Promise<TestResult> {
  const output = await exec(\`npm run \${scope}\`);
  return parseOutput(output);
}
```

### 2. Available Test Commands (cht-core)

| Command | What it runs |
|---------|-------------|
| \`npm run unit-api\` | API unit tests (with nyc coverage) |
| \`npm run unit-sentinel\` | Sentinel unit tests (with nyc coverage) |
| \`npm run unit-webapp\` | Webapp unit tests (Mocha + Karma) |
| \`npm run unit-shared-lib\` | All shared-libs |
| \`npm run integration-api\` | API integration (starts/stops CouchDB) |
| \`npm run wdio-local\` | E2E tests (builds Docker first) |

### 3. Output Parsing

- Parse Mocha \`spec\` reporter output from stdout
- Integration tests also write to \`tests/results/results.txt\`
- nyc \`text-summary\` for coverage

### 4. No Custom Agent Needed

This is a thin wrapper that shells out to npm scripts and parses output. No LLM involved.

### 5. Future Enhancements

- **CTRF**: Structured JSON test results via \`mocha-ctrf-json-reporter\` (high value, low-medium effort for Mocha/WebdriverIO)
- **dependency-cruiser**: Test impact analysis to run only affected tests (medium value, medium-high effort)
- **nyc-diff**: Changed-line coverage reporting (medium value, low effort)

## Dependencies

- #18 (Discovery: Test Orchestration Layer) — tool evaluation and findings

## Enables

- #64 (QA Supervisor) — needs this layer to function

## Complexity

Low

aniket866:
hi @sugat009 
if this issue is  ready to be worked on
PLEASE FEEL FREE TO ASSIGN

Issue #66, opened 2026-03-03
Implement Test Environment Layer


sugat009:
## Summary

Implement the Test Environment Layer that sets up test environments for manual QA testing, based on discovery findings from #16.

## Background

The Test Environment Layer acts as an independent QA engineer, orchestrating environments, preparing config-driven test data, and managing environment lifecycle.

### 1. Environment Orchestration

- Spin up CHT with correct version/branch using \`npm run local-images\` + \`docker compose\`
- Manage lifecycle (start, reset, teardown)

### 2. Config Discovery

- Fetch deployed config from \`GET /api/v1/settings\` (with auth)
- Parse \`contact_types\` for hierarchy structure, \`roles\` for user types, available forms

### 3. Config-Driven Test Data Preparation

- Generate data that conforms to deployed config
- Create valid hierarchies, people, reports
- Override only when ticket explicitly specifies

### 4. Tools

| Tool | Purpose |
|------|---------|
| \`npm run local-images\` | Build local code into Docker images |
| \`docker compose\` | Container orchestration |
| CHT API | Config discovery, data operations |
| \`cht-conf\` | Configuration upload, user creation |
| CouchDB API | Data verification |

### 5. No Custom Agent Needed

This is orchestration of existing tools. No LLM involved in environment setup.

## Dependencies

- #16 (Discovery: Test Environment Layer) — tool evaluation and findings

## Enables

- #64 (QA Supervisor) — needs this layer to function

## Complexity

Medium

Issue #64, opened 2026-03-03
Create QA Supervisor


sugat009:
## Summary

Create the QA Supervisor that orchestrates code validation, test environment setup, and test execution. This is the QA counterpart to #8 (Development Supervisor).

## Background

After the Development Supervisor generates code and tests, the QA Supervisor takes over to validate quality and run tests. It manages three layers:

1. **Code Validation Layer** (#65) — static analysis (ESLint, TypeScript, Prettier)
2. **Test Environment Layer** (#66) — environment setup (Docker, test data, config)
3. **Test Orchestration Layer** (#67) — test execution and coverage (npm scripts, Mocha, nyc)

### 1. QA Workflow (LangGraph)

```
Development Supervisor Output (code + tests)
    ↓
Code Validation Layer (lint, typecheck, format)
    ↓ (blocking failures stop here)
Test Environment Layer (spin up CHT, create test data)
    ↓
Test Orchestration Layer (run tests, collect coverage)
    ↓
Human Checkpoint #2 (review results)
```

### 2. Input/Output

**Input**: Generated code files and test files from Development Supervisor
**Output**: Validation results, test results, coverage report, pass/fail status

### 3. Error Handling

- Validation failures: categorize as blocking (lint errors) vs warning (style issues)
- Test failures: report with details for Human Checkpoint #2
- Environment failures: report setup issues, suggest fixes

## Dependencies

- #65 (Implement Code Validation Layer)
- #66 (Implement Test Environment Layer)
- #67 (Implement Test Orchestration Layer)

## Enables

- Master Supervisor integration
- Human Checkpoint #2

## Complexity

Medium

Issue #63, opened 2026-03-03
Implement Test Generation Layer


sugat009:
## Summary

Implement the Test Generation Layer that generates tests following CHT conventions, based on discovery findings from #15.

## Background

Test generation is deliberately **separate from code generation** so the model doesn't "defend" its own code and skip negative tests. The layer uses CHT test patterns from agent-memory to generate Mocha/Chai/Sinon tests.

### 1. CHT Pattern Awareness

The layer should generate tests using established CHT patterns:
- **Unit**: Mocha + Chai + Sinon, `*.spec.js` naming, `sinon.restore()` in `afterEach`
- **Integration**: Rosie factories, CHT contact hierarchy, `saveDocs()`/`createUsers()` utilities
- **E2E**: WebdriverIO + Page Object Model, `test-id` selectors

### 2. Requirements-Mapped Output

Produce a checklist mapping requirements to test scenarios (happy path, error, edge case, boundary) for human review at Checkpoint #2.

### 3. Pipeline

```
Requirements + Generated Code + CHT Patterns (agent-memory)
    ↓
Test Generation (LLM generates Mocha/Chai/Sinon tests)
    ↓
Basic Gate (compile + pass + increase coverage)
    ↓
Requirements Checklist (requirement → test scenario mapping)
    ↓
Human Checkpoint #2
```

### 4. Future Enhancements

- StrykerJS mutation testing as an internal quality gate
- Tiered code visibility (interface-only for unit, full code for integration)
- TDD mode (test-first generation)

## Dependencies

- #15 (Discovery: Test Generation Layer) — patterns and tool findings

## Enables

- #8 (Development Supervisor) — needs this layer to function

## Complexity

Medium-High

PR #86, opened 2026-04-08
feat(#62): implement code generation layer

### LINKED ISSUE #62: Implement Code Generation Layer
## Summary

Implement the Code Generation Layer with a pluggable module interface, based on discovery findings from #14.

## Background

The Code Generation Layer generates code following CHT patterns and standards. It uses a module interface so different tools (Claude API, Claude Code CLI, OpenCode, etc.) can be plugged in.

### 1. Module Interface

Based on the specification from #14:

```typescript
interface CodeGenModule {
  name: string;
  version: string;
  generate(input: CodeGenModuleInput): Promise<CodeGenModuleOutput>;
  validate?(): Promise<boolean>;
}
```

### 2. Directory Structure

```
src/layers/code-gen/
├── interface.ts
├── modules/
│   ├── claude-code-cli/
│   ├── claude-api/
│   └── opencode/
└── registry.ts
```

### 3. Module Registry

Registry pattern for selecting and loading modules. Extend the existing `LLM_PROVIDER` env variable approach.

### 4. At Least One Working Module

Implement the Claude API module as the default, since it already partially exists.

## Dependencies

- #14 (Discovery: Code Generation Layer) — interface specification

## Enables

- #8 (Development Supervisor) — needs this layer to function

## Complexity

Medium

sugat009:
## Summary

- Add multi-round tool-use loop in Anthropic provider for filesystem access during code generation (read_file, list_directory)
- Add Beads session tracking for code gen pipeline (plans, per-file progress, failures)
- Add selective regeneration: carry forward passing files on retry, only regenerate failing ones
- Replace naive positional diff with LCS-based `diff` package for validation prompts
- Loosen manifest scope so the plan LLM can reference files outside the pre-gathered index (e.g. permission configs, shared settings)
- Exclude test files from code gen plan (handled by separate Test Environment Agent)
- Add per-file validation feedback (FileValidationFeedback) for targeted retries
- Fix disableTools/tools option contradiction across CLI and API providers
- Fix operator precedence bug in looksLikePythonScript
- Extract shared utilities: createModel helper, FailingFileRef type, extractTextContent, capMaxTokens

#62

## Test plan

- [x] `npm run build` passes
- [x] `npm test` passes (291 tests)
- [x] Full workflow tested with Claude CLI provider — generates correct files, no test files in plan
- [x] Full workflow tested with Claude API provider — generates correct files with tool-use, no test files in plan
- [x] Both providers produce functionally equivalent output for the same ticket
- [x] Manifest scope fix confirmed: `default-permissions.js` now included in plan when permission-related features are generated
- [x] Manual verification of generated CHT-Core code on localhost:5988

## Update: legacy test-environment-agent removed

Drops the legacy `src/agents/test-environment-agent.ts` prototype that conflated test generation (#63), test environment setup (#66), and test orchestration (#67) into one LLM-driven agent. The architecture splits these across three layers under a future QA Supervisor (#64). The Development Supervisor's validation node is now code-only; the persistence helpers (`writeToStaging`, `writeToChtCore`, `getAllGeneratedFiles`) no longer reach into `state.testEnvironment`; and the `setupTests` graph node is gone, so the workflow runs `generateCode → validateImpl` directly. This supersedes the "handled by separate Test Environment Agent" note in the summary above.

Commits:
- `b461140` remove legacy test-environment-agent ahead of #63 implementation
- `097f078` drop test-env consumers from validation node and persistence
- `5057169` clean stale test-env comments in edited files

Gates after removal: build and lint green, 629 tests passing, coverage 83.31/69.36/82.11/85.07 (thresholds 78/60/72/78), SonarCloud Quality Gate OK with 0 new issues. QA verified the rewired graph executes at runtime (`generateCode → validateImpl`, no `setupTests`) with no validation-score regression.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sugat009:
Fixing merge conflicts...

Hareet:
@witash 
Could we get your feedback/review on this?

witash (approved):
Its maybe a moot point for this PR, but in future I think we need to split these large PRs into smaller pieces; its not realistic to meaningfully do a line-by-line review of 100 files and 22,000 lines; even if it ends up being mutliple PRs with the same total changes, its easier to understand independent pieces individually.

ran manually over both the sample tickets in the repo (just to see it working) and over the datasource migration tickets.
the datasource migration tickets can't take advantage of the context database as much, and there are some inaccuracies (or maybe preferences where the AI has chosen a different preference) in the ticket descriptions, that lead to inaccuracies in the generated code, but the agent generated apparently working and correct code for each of the tickets, based off the descriptions. Although not completely read to just submit as a PR because of test failures.

a few notes:
this branch implements both beads and has its own todo-tracker.
Are both really necessary?

I looked at the prompt rules, static validators, and the LLM validation.
Also due to the large diff size, its a little hard to understand and evaluate every piece, but not many notes...
For typescript shared-libs, the compile step is important; without it, it does generate code that doesn't compile. But its only for `claude-code-cli`, otherwise it doesn't run compile, is there a reason?

The tickets I tested did not do much on the frontend, so the Angular checks I just skimmed; claude proposed some small tweaks but I don't think its worth it to fiddle with them too much now. I guess these will iterate after this gets merged and gets more usage anyway.

The AST signature validator detect only growth in the parameters, and is relevant only for JS files or when the compile gate is not used, because tsc would catch signature mismatches usually. Most of the cht code is not TS so that's fine, but then do we want to make the signature validator stricter?

One thing thats notably missing is running unit tests...I know that's another PR, but a big difference between using these agents and claude code in interactive mode is that claude code runs the unit tests frequently; for tickets like the datasource migrations which include a lot of modifications and not many additions, tests are a much more concrete and measurable way to validate that nothing existing broke accidentally than the validations in this branch. And small changes to tests can end up being a large portion, even the majority, of the diff. Not running them in the validation stage seems missing, and can generate code that doesn't pass tests.
So nothing to change here, but just noting that when the test branch is merged, I think it will improve the validation as well.

Hareet:
Thanks for some great feedback! @witash I'm concerned about the 2nd paragraph -- Let's fix our data-labeling (domains, descriptions, input tickets, context funneling). Some narrower ideas below

> an manually over both the sample tickets in the repo (just to see it working) and over the datasource migration tickets.
the datasource migration tickets can't take advantage of the context database as much, and there are some inaccuracies (or maybe preferences where the AI has chosen a different preference) in the ticket descriptions, that lead to inaccuracies in the generated code, but the agent generated apparently working and correct code for each of the tickets, based off the descriptions. Although not completely read to just submit as a PR because of test failures.

The datasource tickets struggle because it seems like a domain / context-selection gap: they route to data-sync, whose corpus is all replication memory, not the cht-datasource query-layer shape (qualifier/adapter/endpoint) these tickets would be better off with. The agent injects the wrong pattern there. Root cause is that cht-datasource has no clean home in our 9 domains (pitfall says data-sync, precedent says contacts, template says data-sync). #11174 shows the real axis is a Person/Place/Contact/Report data-access hierarchy. 

Proposing a first-class data-access domain + populating its component index/patterns; I'll open a follow-up and link it here. Seems the description in #11174's "same qualifier union as getUuidsPage," pushed the agent toward the rejected design. cc: @sugat009

sugat009:
@witash thanks for the thorough review, the datasource-migration runs were the most useful signal.

**beads vs todo-tracker:** different layers, not redundant. `TodoTracker` is ephemeral console progress; `BeadsCodeGenSession` wraps the external `bd` CLI for durable issues and is env-gated (a no-op when uninitialized). Removed the unused surface on both.

**Compile gate only for `claude-code-cli`:** closed this cycle, the `claude-api` path now runs the same gate. Since it synthesizes in memory, it snapshots cht-core, materializes the files behind a path guard, runs `tsc --noEmit`, and rolls back. Both Claude transports now type-check their output.

**Stricter AST signature validator:** deliberately light (param-growth only), and with both paths now on the compile gate `tsc` supersedes it, so its niche shrank further. Also fixed a ReDoS in its permission-literal regex (the brittle-regex bit you flagged). TS-aware checks are a follow-up if ever needed.

**Running unit tests in validation:** agreed it matters, but it's the downstream QA phase's job (Test Environment / Orchestration / QA Supervisor), which runs after development and isn't wired in yet. This PR is code-gen; its validation is generation-time (compile + static/AST). Your point is exactly what that phase is for.

Since your review: removed ~470 lines of dead code, fixed a parser bug, added the API compile gate, raised coverage toward main (85/72/84/86), and cleared the SonarCloud findings.

On PR size, fully taken, we'll keep the follow-ups (coverage to 95/85, `writeToChtCore` hardening, the manual `claude-api` run) as separate, smaller pieces.

sugat009:
@Hareet agreed, we settled this in the team meeting: proceed with a first-class datasource/data-access domain, go ahead and raise the PR for it.

Populate it by running the #119 pipeline over the cht-datasource tickets. That lines up with Kombo's observability, once the Langfuse instrumentation (#127) merges and points at the deployed instance, the datasource run rides on the same pipeline pass, so we get traces for it and a clean before/after. And agreed the input labeling (the #11174 description that steered the agent) needs fixing alongside the domain, not the domain alone.

PR #98, opened 2026-04-17
fix(ci): prevent script injection in conventional commits workflow

sugat009:
## Summary
- **Fix SonarCloud security blocker**: PR title was interpolated directly in a `run:` block, enabling shell injection via crafted PR titles. Now passed through an `env:` variable instead.
- **Add `--ignore-scripts`** to `npm ci` to prevent execution of arbitrary post-install scripts.
- **Update actions** (`checkout@v6`, `setup-node@v6`) and Node (`22.x`) to match the rest of the CI.


Hareet (approved):
LGTM!

PR #120, opened 2026-06-24
chore(memory): promote strong-fit messaging drafts from memory-pipeline for review

Hareet:
Promotes **17 strong-fit `messaging` drafts** from `agent-memory/_pending/` into `agent-memory/domains/messaging/issues/` for squad content review.

**Categories:** bug (10), feature (7)
**Themes:** SMS gateway integrations (Africa's Talking, Nepal centralized gateway, RapidPro), scheduled-task/message-state handling, mustache phone helpers.

All 17 carry `domainFit: strong` + a `## Domain Rationale` section. Weak-fit messaging drafts: 0 (none deferred).

Merge order: **#1 of 9** (messaging → infrastructure → forms-and-reports → tasks-and-targets → authentication → contacts → interoperability → configuration → data-sync).


Hareet:
@sugat009 Needs to be rebased with main after #119 is merged

sugat009:
Heads up, this one currently has merge conflicts with its base and will need a rebase before it can merge.

agent-memory/domains/messaging/issues/10082-fix10068-fix-africas-talking-request-body-parsing-10073.md
```diff
@@ -0,0 +1,92 @@
+---
+id: cht-core-10082
+category: bug
+domain: messaging
+domainFit: strong
+issueNumber: 10082
```
sugat009: **issue:** `issueNumber: 10082` is the PR number; PR #10082 closes issue **#10068** (so does #10073). Both drafts should key to 10068 and dedup into one memory.

sugat009 (changes requested):
Content review (domain fit + distillation accuracy). The distilled prose is faithful and reusable; I verified several drafts against their real cht-core PRs (9559, 10868, 10945) and found no hallucinated behavior. The blockers are identity and structure:

- **issue (blocking): identity keys record the PR number, not the issue, on 12 of 17 drafts.** `id`/`issueNumber`/`issueUrl` point at the source PR rather than the issue it closes (e.g. `10082` keys to 10082, but PR #10082 closes issue #10068). Verified against `closingIssuesReferences`. Invisible to `validate-schema` because `/issues/N` redirects to `/pull/N`.
- **issue (blocking): 3 duplicate clusters (6 files).** issue 10068 = [10073, 10082], issue 10225 = [10230, 10243], issue 10802 = [10803, 10811]. Each is a base PR plus a backport/sibling distilled as a separate record; they should collapse to one memory per issue with a `source_prs[]` array.
- **issue: `related_issues: []` is empty on every draft** though the prose "## Related Issues" is populated. The machine-readable linkage was never backfilled.
- **issue (content): `10729-smsparser-bugs-typos.md`** (a pre-existing draft this PR only modifies) is distilled from **issue** #10729, not a merged PR; its Solution/Testing sections present the issue's *proposed* fixes as completed and tested. The real work merged as PR #10730. Worth a corpus fix.
- **issue: `domainFit: strong` on every draft (0 weak).** `8717` (add link to sender contact page) is a clear example: its files are entirely `webapp/.../sender.component.*` UI navigation, and its own `related_workflows` is `ui-extensions`. Milder borderline picks: `8774`, `10853`, `10477`.
- **praise:** prose fidelity is genuinely high (RapidPro 400/200 handling in 9559, the `local_phone` template syntax in 10868 both reproduced exactly).

Clean bill: no secrets, no PII, schema 100% valid.

agent-memory/domains/messaging/issues/10802-message-pending-state.md
```diff
@@ -6,7 +6,7 @@ subDomain: scheduled-tasks
 issueNumber: 10802
```
sugat009, line 6: **issue (accuracy):** this draft (Code Patterns / Solution, ~L54) uses `task.status`, but the real scheduled-task field is `task.state` (`SCHEDULED_STATE='scheduled'` in `shared-libs/transitions/src/schedule/due_tasks.js`). The `isDue()`/`due_date` snippet doesn't exist (the code uses `task.due || task.timestamp || doc.reported_date`), and the Related Files list `api/src/controllers/scheduled_tasks.js` + `sentinel/src/schedule/task-processor.js`, which neither source PR (#10803/#10811) touches. For a memory draft these would mislead an agent — please align to the real code or drop the fabricated specifics.

Hareet:
Re-review prep, same protocol as the configuration batch (#130): both round-2 items are fixed, and I then re-verified every factual claim in all 17 drafts against the cht-core commit each describes — first with the grounding prober (extended again, see tally), then by hand for what probes can't see. Fixes landed in d92041d (grounding pass) and 30f0560 + 4dfb161 + c0237f4 (this round); `validate-schema` passes 76/0. Each point carries the command that establishes it.

### The two review items

**10802 — fabricated detail replaced with the real guard.** `task.status`/`isDue()`/`due_date` are gone; the draft now carries the actual mechanism, verified line-for-line on master:

```sh
git -C $CORE grep -nE "SCHEDULED_STATE = |task.state !== |task.due \|\|" origin/master -- shared-libs/transitions/src/schedule/due_tasks.js
# :14 const SCHEDULED_STATE = 'scheduled';   :50 if (task.state !== SCHEDULED_STATE) {   :55 let due = task.due || task.timestamp || doc.reported_date;
```

The two Related Files neither source PR touches are dropped, and the two-PR story is now precise — the patches are byte-identical and each line carries exactly one:

```sh
diff <(git -C $CORE show 6a5867bb --format= -- shared-libs/) <(git -C $CORE show b87a025f --format= -- shared-libs/) && echo IDENTICAL
git -C $CORE branch -r --contains 6a5867bb | grep -E 'origin/(master|5\..\.x)$'   # master + 5.2.x
git -C $CORE branch -r --contains b87a025f | grep -E 'origin/(master|5\..\.x)$'   # 5.1.x only
```

**10868 — classifier scaffolding stripped**: "…so the configuration pitfall does not apply" is gone; the rationale states the fit without rubric vocabulary.

### What full re-verification found beyond the two items

1. **Round-1's identity defect survived in the two hand-authored drafts.** `4278` and `8492` still keyed `id`/`issueNumber`/`issueUrl` to PR numbers — the relink tool reads machine frontmatter, and these pre-existing files never got the treatment. Their own prose already named the real issues; frontmatter now agrees (4278 → **#3738**, 8492 → **#8414**), and both gained machine anchors (`source_prs` + the PR's merge commit as `source_sha`), so grounding checks them at their own trees instead of guessing against master:

```sh
curl -s https://api.github.com/repos/medic/cht-core/pulls/4278 | jq -r '.body' | head -1   # "Issue: #3738"
curl -s https://api.github.com/repos/medic/cht-core/pulls/8492 | jq -r '.title'           # fix(#8414): sms gateway test flakiness
```

2. **The newly anchored probes then caught three residual claims my grounding pass had missed.** In 10802: "Filter tasks by both `due_date` and `status` fields" (neither field exists — same fabrication family as your inline, one bullet over) and "Added a sentinel integration test" for a spec file that pre-existed (6a5867bb *adds a 69-line case* to it: M, not A) — corrected in 4dfb161. In 9467: an "`api/v2/broadcast`" endpoint the service does not call — it posts to `/api/v2/broadcasts.json` — corrected in c0237f4. All three only became checkable once their drafts anchored.

```sh
git -C $CORE show 6a5867bb --stat -- tests/integration/sentinel/schedules/due-tasks.spec.js   # 1 file changed, 69 insertions(+)
git -C $CORE grep -n "broadcasts.json" da4b50f7 -- api/src/services/rapidpro.js              # :92  requestOptions.uri = '/api/v2/broadcasts.json';
```

3. **10442 ↔ 10802 are an in-batch pair** — both rework `due_tasks.js` state transitions — but only 10802 carried the back-reference. `related_issues` on 10442 now links `cht-core-10802`, making the pair symmetric (the 9696/9727 class from the configuration batch).

4. **The deep rewrites in d92041d spot-verify exactly.** 10073's inversion fix (the bug was outbound *response* parsing, not inbound requests) matches fb52de8606: `parseResponseBody` removal, `.then(result =>`, the 6-insertion/6-deletion fixture change, the `(CREDENTIAL_PASS_OUTGOING ? describe : describe.skip)` gate keyed on `AFRICAS_TALKING_SANDBOX_API_KEY`, and the incidental `getDoc(id, rev = '', ...)` default. 10442's snippets are character-exact at 862f69a6 (down to the `hasValidMessage?` spacing). 10729 is re-grounded on merged PR #10730 per your round-1 note, with line-precise spec claims that hold (`list:` fixtures at 483/490/508/527/568, all `type: 'integer'`; 393b9d6a touches no test files).

### Prober tally

Tooling first: my July-27 run left 24 claims unverifiable across five drafts (unresolvable anchors) and reported 5 defects. All five were probe blind spots, not draft errors, fixed on #145 (adfb256 + 18784de): hand-authored `source_prs[]` anchors now resolve as canonical, line-wrapped member chains (`request\n  .post({` at africas-talking.js:80–81) match, path absence degrades by provenance instead of manufacturing defects, and unstamped merges resolve via the GitHub API — 9467 anchors at da4b50f7 ("fix(#9467): better handling of RapidPro error codes") exactly that way.

Re-running `ground-claims` over this branch's 17 drafts with those fixes (cht-core @ `1f3e56f226`): **217 grounded / 2 ungrounded / 0 unverifiable / 0 anchor-unusable — all 17 drafts anchored at their own commits** (my July-27 run: 189 / 5 / 24 with five drafts unanchored). Claim extraction is LLM-driven, so the extracted set varies slightly run to run; every flag across three runs traced to a probe or extraction artifact rather than a draft error, and the two current ones adjudicate cleanly:

- **10442 `sms.clear_failing_schedules`** — an app-settings key never appears as a dotted token in code; the setting is real, spelled through the accessor:

```sh
git -C $CORE grep -n "clear_failing_schedules" 862f69a6 -- shared-libs/transitions/src/schedule/due_tasks.js
# const clearFailing = config.get('sms')?.clear_failing_schedules || false;
```

- **10073 `parseResponseBody`** — extraction read a *removal* claim ("Removed the `parseResponseBody` helper…") as an existence claim; the anchor tree correctly lacks the symbol because the PR deleted it:

```sh
git -C $CORE show fb52de8606 -- api/src/services/africas-talking.js | grep '^-.*parseResponseBody'
# -const parseResponseBody = body => {
# -      const result = parseResponseBody(body);
```

To reproduce the run (the prober lives on #145's branch; drafts on this one):

```sh
git worktree add /tmp/msg memory/promote-messaging && git checkout memory/draft-verification
CHT_CORE_PATH=<cht-core clone> LLM_PROVIDER=claude-cli \
  npm run ground-claims -- --dir /tmp/msg/agent-memory/domains/messaging/issues --label recheck
```

**Cross-cutting, unchanged from your note:** filenames still encode PR numbers, the two draft schemas still coexist, and the five untouched pre-existing messaging drafts (3406/4039/4374/6995/7105) share the hand-authored anchor gaps — all corpus-cleanup candidates outside this PR's diff.

agent-memory/domains/messaging/issues/10802-message-pending-state.md
```diff
@@ -34,29 +39,38 @@ The scheduled task processing logic in `shared-libs/transitions/src/schedule/due
 2. When a document had multiple `scheduled_tasks` with identical due dates and one was stuck in `scheduled` state, the system continuously added already-processed messages back to the pending queue
 3. Created a feedback loop where the same message was resent every processing cycle (every 5 minutes) regardless of its current status
 
-The code only checked `due_date` without checking if the task was already processed or in a different state.
+The code only checked the task's computed due value (`task.due || task.timestamp || doc.reported_date`) against the collected due dates, without checking whether the task was already processed or had moved to a different `state`. More precisely, `due_tasks.js` trusted the (eventually-consistent) `messages_by_state` CouchDB view results and did not re-verify each message's current state on the freshly loaded/hydrated document before mutating it; a message already transitioned out of the schedulable state (muted, cleared, or already sent) still appeared in the view and was reprocessed (PR #10803, PR #10811).
```
sugat009: issue (blocking): the second sentence states a mechanism the code rules out. A message that had already transitioned never appears in the view result: `execute()` queries `messages_by_state` keyed on `['scheduled', due]`, so the view returns only the still-scheduled sibling task. `processBatch` collects that row's due date into the per-document `dueDates` list, and `updateScheduledTasks` then loops over all of `doc.scheduled_tasks`, promoting any task that matches on due date alone. That loop is what reached the already-sent message, and the bug reproduces with a perfectly fresh index, so eventual consistency points at the wrong fix. Bullets 1-3 above already state the real mechanism correctly.

```suggestion
The code only checked the task's computed due value (`task.due || task.timestamp || doc.reported_date`) against the collected due dates, without checking whether the task had moved to a different `state`. More precisely, the view vouches for one task while the update loop acts on all of them: `execute()` queries `messages_by_state` with `startkey`/`endkey` keyed on `['scheduled', due]`, so only still-scheduled tasks are returned; `processBatch` groups those rows by document and collects their due values into a single per-document `dueDates` list; and `updateScheduledTasks` then iterates every entry in `doc.scheduled_tasks`, promoting any whose computed due value appears in that list. A task that had already moved out of `scheduled` was therefore never returned by the view at all. It was reached because it shared a due date with a still-scheduled sibling on the same document, which is why the fix is a per-task state guard inside that loop rather than a change to the view or the query (PR #10803, PR #10811).
```

agent-memory/domains/messaging/issues/10802-message-pending-state.md
```diff
@@ -73,15 +87,14 @@ Chose to fix at the library level (`shared-libs/transitions`) rather than in ind
...
 
 ## Testing
 
 - Added integration test to verify that only scheduled tasks are processed in each window
 - Test simulates multiple tasks with same due date and different statuses
 - Verified that processed tasks don't get re-added to pending queue
-- Edge case testing for tasks with null/undefined status fields
+- Added a sentinel integration test case (+69 lines) to the pre-existing `tests/integration/sentinel/schedules/due-tasks.spec.js` asserting scheduled_tasks are only updated to `pending` when still in the `scheduled` state and left untouched otherwise (PR #10803, PR #10811)
 
 ## Related Issues
```
sugat009, line 101: issue (blocking): line 102 cross-references #10754 as 'Scheduled task duplicate processing (similar issue)'. cht-core #10754 is 'Cookies not being sent with `secure: true`', a closed cookie bug in api, unrelated to scheduled tasks. Please delete the bullet or replace it with a genuinely related issue.

agent-memory/domains/messaging/issues/10073-fix10068-fix-africas-talking-request-body-parsing.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Related Issues
+
+- #10068: Africa's Talking integration tries to parse an already-parsed request body
+
+## Domain Rationale
+
+**Fit:** strong
+
+Africa's Talking is an SMS gateway CHT uses to send and receive messages, and the bug is in the inbound request-parsing code of that messaging transport service. The build.yml change only wires the new e2e test into CI, so the PR's substance is messaging behavior, not infrastructure or gateway setup/config.
```
sugat009: issue (blocking): this contradicts the Root Cause and Code Patterns above, which now state the bug was outbound response parsing and that nothing inbound is involved anywhere in the file. The inbound webhook side lives in `api/src/controllers/africas-talking.js`, which PR #10073 never touched, so this sentence points an agent at a real but wrong file.

```suggestion
Africa's Talking is an SMS gateway CHT uses to send and receive messages, and the bug is in the outbound send path of that messaging transport service: the service double-parsed the response body of its own POST to the gateway. The build.yml change only wires the new e2e test into CI, so the PR's substance is messaging behavior, not infrastructure or gateway setup/config.
```

agent-memory/domains/messaging/issues/10073-fix10068-fix-africas-talking-request-body-parsing.md
```diff
@@ -0,0 +1,89 @@
...
+- tests/e2e/default/sms/africas-talking.wdio-spec.js
+- tests/utils/index.js
+
+## Testing
+
+Updated unit-test fixtures in api/tests/mocha/services/africas-talking.spec.js (the three stringified response fixtures became plain objects; 6 insertions / 6 deletions, no new cases) and added a CI-only e2e case to tests/e2e/default/sms/africas-talking.wdio-spec.js that interacts with the Africa's Talking Sandbox server. The case is a `(CREDENTIAL_PASS_OUTGOING ? describe : describe.skip)` block keyed on `process.env.AFRICAS_TALKING_SANDBOX_API_KEY`, which .github/workflows/build.yml supplies from repository secrets; the tests/utils/index.js edit in the same commit is an incidental `getDoc(id, rev = '', ...)` default and does not support the e2e test. The fix was also confirmed by manual testing.
+
+## Related Issues
+
+- #10068: Africa's Talking integration tries to parse an already-parsed request body
```
sugat009: issue: 'request body' keeps the pre-grounding framing, and the gloss does not match the issue. #10068 is titled 'CHT SMS message sent via Africa's Talking stuck in "pending" state', and the double-parse was of the response body.

```suggestion
- #10068: CHT SMS message sent via Africa's Talking stuck in "pending" state (the service re-parsed the already-parsed response body of its own send request)
```

agent-memory/domains/messaging/issues/10073-fix10068-fix-africas-talking-request-body-parsing.md
```diff
@@ -0,0 +1,89 @@
...
+issueUrl: https://github.com/medic/cht-core/issues/10068
+title: Fix Africa's Talking SMS gateway double-parsing of the outbound send response body
+lastUpdated: '2026-07-16'
+summary: The Africa's Talking SMS integration JSON-parsed the response body of its own outbound send request, which `@medic/couch-request` had already parsed, so every send failed to produce a message state change. The fix removes the redundant `parseResponseBody` helper and appends a CI-only e2e case (gated on `AFRICAS_TALKING_SANDBOX_API_KEY`) to the pre-existing Africa's Talking wdio spec, exercising the Sandbox server.
+services:
+  - api
+techStack:
+  - javascript
+  - nodejs
+  - express
```
sugat009: nitpick: the Root Cause now states no Express middleware is involved anywhere in this file, so `express` here keeps the old framing.

```suggestion
```

agent-memory/domains/messaging/issues/9467-rapidpro-error-handling.md
```diff
@@ -21,33 +23,37 @@ When RapidPro SMS gateway responded with a 400 error for invalid phone numbers (
 
 1. Continuously retried sending the same message over and over
 2. Threw `StatusCodeError: 400 - {"urns":{"0":["Invalid URN: tel:********. Ensure phone numbers contain country codes."]}}`
-3. Then crashed with `TypeError: Cannot read property 'state' of undefined` at `getStateUpdate()` in `rapidpro.js:79`
+3. Separately, on a 200 whose body carried no (or an unmapped) `status`, `remoteStatusToLocalState()` returned `undefined` and `getStateUpdate()` threw `TypeError: Cannot read property 'state' of undefined` at `rapidpro.js:79` (`state: status.state`). This is a different code path from the 400 above, which rejects into `.catch` and never reaches line 79.
 
 This caused message flooding and filled logs with repeated errors. One deployment reported needing to manually clear 62,000 scheduled messages.
```
sugat009: issue (blocking): the 62,000 figure belongs to #10428, not to this bug. Its body reads 'Recently, we had to clear 62,000 scheduled messages from the past in one instance', and the context there is the empty-message workaround for nonexistent phone numbers leaving messages in scheduled state forever, a different mechanism from the 400 retry loop. Placed here it reads as a consequence of this bug.

```suggestion
This caused message flooding and filled logs with repeated errors.
```

agent-memory/domains/messaging/issues/9467-rapidpro-error-handling.md
```diff
@@ -21,33 +23,37 @@ When RapidPro SMS gateway responded with a 400 error for invalid phone numbers (
...
 ## Solution
 
 Updated the RapidPro error handling to:
 
-1. Catch 400-level HTTP errors from RapidPro API
-2. Parse the error response structure (which contains `urns` validation errors)
-3. Mark the message as `failed` instead of leaving it in a retryable state
-4. Log the validation error clearly for debugging
-5. Prevent further retry attempts for messages with invalid recipient phone numbers
+1. In `sendMessage`'s existing `.catch`, test exactly `err?.statusCode === 400` (the added comment cites https://rapidpro.io/api/v2/ — "Do not retry with the same values")
```
sugat009: suggestion (non-blocking): time-scope the property name. Current master reads `err?.status === 400` (api/src/services/rapidpro.js:109), so `statusCode` is the PR #9559 era shape, and the Testing bullet's `.rejects({ statusCode: 400 })` stub has the same vintage. A parenthetical like the one the 10604 draft uses ('at the time of this fix ...') keeps it accurate against today's source.

agent-memory/domains/messaging/issues/4278-sms-gateway-api-tests.md
```diff
@@ -27,7 +30,7 @@ The SMS gateway API endpoint had no test coverage, making it risky to modify or
 
 ## Root Cause
 
-The SMS gateway endpoint (`api/src/controllers/sms-gateway.js`) was implemented without corresponding tests. This violated testing best practices and made the codebase fragile.
+The SMS gateway endpoint (`api/controllers/sms-gateway.js` — the `api/src/` layout postdates this change) was not untested: `api/tests/unit/controllers/sms-gateway.js` covered it at the unit level, and `tests/protractor/e2e/sms-gateway.js` already POSTed to `/api/sms` as part of browser-driven UI assertions. What was missing was a dedicated API-level e2e suite asserting the `/api/sms` request/response contract and the resulting database state on their own, which made the endpoint risky to modify or extend.
```
sugat009, line 34: issue: the Problem section (line 23) still opens with 'The SMS gateway API endpoint had no test coverage', which this corrected Root Cause explicitly refutes. Line 23 sits outside this diff so I cannot attach a suggestion there; proposed replacement for that sentence: 'The SMS gateway API endpoint had unit coverage and incidental browser-driven coverage, but no dedicated API-level e2e suite asserting the request/response contract and resulting database state, making it risky to modify or extend.'

agent-memory/domains/messaging/issues/4278-sms-gateway-api-tests.md
```diff
@@ -41,41 +44,37 @@ Added comprehensive test suite for SMS gateway API endpoint:
      // ... assertion logic
```
sugat009: issue: this snippet is a composite that appears nowhere in the spec. The real `check()` functions (spec lines 191 and 281 at `d88f2e2`) call `getMessageContents()` / `getMessageStates()`, and the only `utils.db.query` in the file lives inside `allMessageDocs()` at line 362, with `{ reduce: false, include_docs: true }`. As written the fence drops `include_docs` and contradicts bullet 3 directly below it. Suggest replacing the fence with the real helper:

```js
function allMessageDocs() {
  return utils.db.query('medic-client/messages_by_contact_date',
      { reduce:false, include_docs:true })
    .then(res => res.rows.map(row => row.doc));
}
```

agent-memory/domains/messaging/issues/4278-sms-gateway-api-tests.md
```diff
@@ -41,41 +44,37 @@ Added comprehensive test suite for SMS gateway API endpoint:
...
    }
    ```
-3. **Used changes listener approach**: Registered CouchDB changes listener to capture documents created by endpoint calls
+3. **Polled a view, not a changes listener**: `allMessageDocs()` reads `medic-client/messages_by_contact_date` with `{ reduce: false, include_docs: true }` and maps `res.rows` to docs; no CouchDB changes listener is used
 4. **Tested message flow**: Verified messages sent to gateway endpoint are properly stored in database
 5. **Validated error handling**: Tested endpoint behavior with invalid content
```
sugat009: issue: the PR's single file contains no invalid-content or error-handling test. The only match for 'invalid' or 'error' in all 365 lines is a fixture field `errors: []`. The same claim repeats in Testing (line 77, 'Verified error handling for invalid content'). Please delete this bullet and that Testing line. Problem item 4 can stay, it only lists a risk area.

agent-memory/domains/messaging/issues/4278-sms-gateway-api-tests.md
```diff
@@ -41,41 +44,37 @@ Added comprehensive test suite for SMS gateway API endpoint:
...
 
 ## Code Patterns
 
 - Use polling with timeout for async test assertions (100ms intervals)
-- Pattern: Register changes listener before API call to capture resulting documents
+- Pattern: after the API call, poll the view in a `check()` loop until the expected snapshot matches or a 10s deadline expires, then resolve/reject
```
sugat009, line 57: suggestion (non-blocking): the `tests/protractor` tree no longer exists on master (e2e suites live under `tests/e2e` since the wdio migration), and lines 54, 56 and 70 present the path as a live pattern. Time-scope it the way the Root Cause already does for `api/src/` ('the `api/src/` layout postdates this change').

agent-memory/domains/messaging/issues/10729-smsparser-bugs-typos.md
```diff
@@ -6,25 +6,29 @@ subDomain: sms-parsing
...
+1. **String list parser bug**: The code used `for (const i of field.list)` which yields array items, not indices. Then it used `field.list[i]` treating `i` as an index, making `item` `undefined`. With the conventional `[[value, label], ...]` list shape the very next line, `item[0] === cleaned`, therefore threw `TypeError: Cannot read properties of undefined (reading '0')`, which propagated out of `smsparser.parse` uncaught — the loop crashed rather than falling through to the "Option not available" warning. That warning was effectively unreachable: it required `field.list` to be empty (or to hold entries that coincidentally stringify to valid indices).
 
 2. **parseArray null crash**: The function called `getParser(def, doc)` before checking if `def` was null. When `def` is null/undefined, `getParser()` returns `undefined`, and calling `parser(def, doc)` threw `TypeError: parser is not a function`, crashing the SMS parsing process.
```
sugat009: issue (blocking): the mechanism is over-broad, and the draft's own Testing section already refutes it: line 75 correctly notes that `getParser(null, doc)` returned `textformsParser.parse` for the junk-data spec case. `getParser` returns `undefined` only for a non-string `doc.message` (smsparser.js:76-78) or an unrecognized Muvuku format code (:86-87), not merely because `def` is null.

```suggestion
2. **parseArray null crash**: `parseArray` called `getParser(def, doc)` and invoked the returned parser before its `if (!def || !def.fields) return []` guard ran. For message shapes where `getParser` returns `undefined` (a non-string `doc.message`, or an unrecognized Muvuku format code), `parser(def, doc)` threw `TypeError: parser is not a function` before the guard could return. `getParser` does not return `undefined` merely because `def` is null; for an ordinary text message it returns `textformsParser.parse`.
```

agent-memory/domains/messaging/issues/10729-smsparser-bugs-typos.md
```diff
@@ -62,15 +68,13 @@ Chose to fix the bugs directly rather than refactoring the entire parsing system
...
 - api/src/controllers/infodoc.js
 - sentinel/src/schedule/reminders.js
-- api/tests/mocha/services/report/smsparser.js
+- api/tests/mocha/services/report/smsparser.spec.js (existing spec; not modified by #10730)
 
 ## Testing
 
-- Bug fixes covered by existing unit tests in `api/tests/mocha/services/report/smsparser.js`
-- No new tests needed as the existing tests already validated the correct behavior
+- PR #10730 (393b9d6a) touched no test files, and neither fix is covered by existing tests. In `api/tests/mocha/services/report/smsparser.spec.js` every `list:` fixture (lines 483, 490, 508, 527, 568) is on a `type: 'integer'` field, so the `string` list branch is untested; and the only `parseArray(null, doc)` case ('junk example data', line 1339) never reached the crash because `getParser(null, doc)` returns `textformsParser.parse` for that message. Note the spec path is `api/tests/mocha/services/report/smsparser.spec.js` — there is no `smsparser.js` under that directory.
```
sugat009, line 76: issue: Design Choices (line 62) still claims 'Existing unit tests already covered the functionality', which this grounded Testing section contradicts ('neither fix is covered by existing tests'). Line 62 sits outside this diff; please delete that bullet.

agent-memory/domains/messaging/issues/10497-feat10459-allow-multiple-sms-recipients.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Related Files
+
+- shared-libs/message-utils/src/index.js
+- shared-libs/message-utils/test/index.js
+- tests/integration/sentinel/schedules/schedules-recipients.spec.js
+
+## Testing
+
+Unit tests added/updated in shared-libs/message-utils/test/index.js for multi-recipient resolution, plus an integration test for sentinel scheduled-message recipients in tests/integration/sentinel/schedules/schedules-recipients.spec.js. Review feedback was limited to minor stylistic comments.
```
sugat009: issue: the last sentence is not accurate. Review feedback included substantive behavioral requests: a redesign of `normalizeRecipient` (coerce to string, trim, filter empties, always return an array), an `(!recipient.length)` correction, removal of a then-redundant section, and handling for non-string truthy values.

```suggestion
Unit tests added/updated in shared-libs/message-utils/test/index.js for multi-recipient resolution, plus an integration test for sentinel scheduled-message recipients in tests/integration/sentinel/schedules/schedules-recipients.spec.js.
```

agent-memory/schema.json
```diff
@@ -138,6 +138,15 @@
...
           "description": "Source PR reference, format: medic/<repo>#<number>"
         },
+        "source_prs": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "pattern": "^medic/(cht-core|cht-interoperability)#[1-9][0-9]*$"
+          },
+          "minItems": 1,
+          "description": "All PR refs that resolve to this issue (backport cherry-picks, multi-PR epics), set by the dedup pass when more than one draft collapsed into this one. source_pr remains the canonical PR."
```
sugat009, line 148: nitpick: five drafts in this PR (10729, 10802, 4278, 8492, 9467) carry `source_prs` with no `source_pr`, while this description says `source_pr` remains the canonical PR. Either add the canonical `source_pr` to those five or relax this sentence.

agent-memory/domains/messaging/issues/10073-fix10068-fix-africas-talking-request-body-parsing.md
```diff
@@ -0,0 +1,89 @@
+---
+id: cht-core-10068
+category: bug
+domain: messaging
+domainFit: strong
+issueNumber: 10068
+issueUrl: https://github.com/medic/cht-core/issues/10068
+title: Fix Africa's Talking SMS gateway double-parsing of the outbound send response body
+lastUpdated: '2026-07-16'
```
sugat009: nitpick: ten drafts substantively rewritten in this iteration keep their pre-rework `lastUpdated` (10073, 10477, 10497, 10729, 10868, 10945, 8717, 8774, 9022, 9364), while the five others touched in the same commits were bumped to '2026-07-28'.

Hareet:
Added a coherence check. Increased count to 3 passes over gates.

All 16 inline items are fixed. Your diagnosis of the shared cause was right, and acting on it as a *class* rather than a list turned up more of the same in drafts you had not flagged — plus a reason to take five drafts back out of this PR, and one correction to my own evidence that I would rather flag than have you find. The two withdrawals are noted; thanks for re-checking them.

Head is `e3672a5`, and the diff is exactly the 17 drafts you reviewed plus `schema.json`.

### The four sections, checked rather than promised

You were exactly right about the mechanism: the grounding pass corrected the sections that assert mechanism against code and left the interpretive ones asserting what it had just disproved. Both halves then ground independently while the document tells two stories, which is why no probe caught it — every existing check compares a draft to something *outside* it, and this one is text against text.

So it is a check now. `check-coherence` sends each draft to a model that may only *identify* contradicting pairs and must quote both sides verbatim; the script then verifies mechanically that both quotes occur in the file and discards any that do not, so a fabricated quote cannot land. Run blind it found your three plus these, in drafts that remain here:

- **10802** — one sentence claimed the fix landed "on master and 5.2.x" *and* that 5.1.x "is the only line carrying it". My wording, from the last round. Each patch reaches a different set of lines; reworded.
- **8717** — Solution credited "supporting navigation logic in `sender.component.ts`", which its own Code Patterns refutes (no `Router` injected, two data accessors, routing is the declarative `routerLink`).
- **8492** — Problem blamed "inconsistent message state setup in test factories" while its Root Cause says the cause was shared mutable fixture state, *not* the factory failing to set a state. A second pair in the same draft then turned up in Design Choices ("proper state setup"), which is why one fix per file is not a finish line.
- **10442** — the worst of them, and mine. See below.
- **10729** — the summary called the string-list bug a silent "fields never match" failure, which its own Problem section refutes: the loop threw `TypeError` on `item[0]` and propagated out uncaught.
- **9364** — Domain Rationale said "raw SMS content" while Code Patterns says the sanitisation is post-tokenisation and explicitly not raw-message preprocessing.

**10442 deserves its own paragraph, because my round-3 sweep reproduced the exact class it was fixing.** I rewrote Problem to the real code path and left `summary` and Design Choices asserting the old story. Ground truth at `862f69a6^`:

```sh
git -C $CORE show 862f69a6^:shared-libs/transitions/src/schedule/due_tasks.js | sed -n '76,80p'
#       // only update task states when messages exist
#       if (task.messages) {
#         updatedTasks = true;
#         utils.setTaskState(task, 'pending');
```

So a task *with* a `messages` array but an empty body was promoted to `pending`; only a task with no array at all sat in `scheduled`. The summary's "they sat in `scheduled` indefinitely" was false for half the cases, and Design Choices' "deployments keep leaving such messages indefinitely scheduled" contradicted the Solution's own note that leaving them in `scheduled` "is itself a change". All four sections now describe one path.

### Your items

**Blocking.** 10802's Root Cause now states the real mechanism (your suggestion, adopted verbatim — I verified the view genuinely cannot return an already-transitioned task: `execute()` keys on `['scheduled', due]` and the map emits `[task.state, due]`). The #10754 bullet is deleted. 10073's Domain Rationale and Related Issues gloss are yours as written, and `express` is out of techStack. 9467 loses the 62,000 figure. 10729's `parseArray`/`getParser` bullet is yours, verified at `smsparser.js:74-76` and the `default:` branch.

**The rest.** 4278's Problem takes your replacement sentence; the composite fence is replaced by the real `allMessageDocs()` (your snippet, verbatim from `d88f2e2:361-365`); both invalid-content claims are gone — the 365-line spec's only `invalid|error` match is a fixture field `errors: []` at :233. 10729's "existing unit tests already covered" bullet is dropped. 10497 no longer calls the feedback stylistic. 9467's `statusCode` and 4278's `tests/protractor` are time-scoped; 4278 gets one note covering all four dead paths, since a drift check found eight unqualified mentions across them rather than the three lines review listed.

A cross-reference audit also found PRs cited in Related Issues as though they were issues, in 4278, 8492 and 9467; all are labelled `PR #N` now. Per your schema note, the canonical `source_pr` is added to the drafts that carried only `source_prs`. And three more corrections fell out of an independent audit of this work: 9467 said a `logger.error` call "was moved" when the diff shows the `// ignore error` *comment* moved and the log line is untouched context; 8492 described PR #6995 as "related testing improvements" when it is "Adds RapidPro as an SMS Gateway"; and 9022 implied it added a Sentinel integration spec when `diff-tree` shows only `tests/e2e/default/reports/sms-messages.wdio-spec.js` was added and the integration spec was extended (+160/−90).

### The other five drafts are moving to their own PR

I did re-key 3406, 4039, 4374, 6995 and 7105 — then reverted it, because giving them anchors made their claims checkable for the first time and several do not survive it:

```sh
git -C $CORE grep -nFw handleCallback e9e305d2ee   # 0 hits
git -C $CORE grep -nFw RAPIDPRO_URL e9e305d2ee     # 0 hits — credentials come from
                                                   # secureSettings.getCredentials('rapidpro:outgoing')
git -C $CORE ls-tree f1ab8b867b -- shared-libs/messaging   # no such path at the anchor
```

6995 names two env vars and a `handleCallback` that do not exist. 4374 names three files as touched that its backport commit never touched. 3406 contradicts itself on view keys — recommending compound keys *and* string keys, while the PR emits only `msg.uuid` and `task.state`.

**One correction to my own evidence.** My first pass counted 19 ungrounded claims across the five and led with "seven fabricated metric names in 7105". That part was wrong. All seven exist at `67626fff` as nested keys of the monitoring response — `backlog:` (:293), `total: jsonV1.messaging.outgoing.state` (:325, literally the rename the draft describes), `seven_days:` (:326), `last_hundred:` (:327) — and that PR is what adds `failed`/`delivered` to the v1 counters. A dotted path like `monitoring.messaging.outgoing.seven_days` describes JSON shape and cannot grep as one token: the same extraction artifact I had correctly dismissed four times elsewhere in this batch, and then booked as evidence. `b78ed35` records that on the branch.

The deferral still stands on 6995, 4374, 4039 and 3406, which do check out — but on less than I first claimed, and if you would rather see the five fixed here than deferred, that is a fair read of a thinner case.

**The follow-up PR will carry:** the five re-keys with anchors (3406→3073, 4039→3627, 4374→4110, 6995→6532, 7105→6572 — all verified real issues with matching titles), the content fixes those anchors demand, the remaining `PR #N` labels, and 4374's reference to the re-keyed 4278 entry, which this PR does make stale at line 78. I left that one line to travel with its own draft rather than reach into a file this cycle deliberately reverted out of scope — say the word if you would rather not merge the stale pointer and I will pull it in here.

Reverting 3406 also removed a contradiction this branch had briefly introduced, worth admitting since it is the same class: my Code Patterns rewrite there was right about the view but I had not read Design Choices, which still claimed the PR emits both key shapes.

### Gate

`validate-schema` 76/0. `verify-drafts --online`: **zero findings across all 17 in-scope drafts** (the five reverted ones report their pre-existing PR-keyed identity, which is main's state and outside this diff). `ground-claims`: **229 grounded, 0 stale-as-written, 0 unmatched code fences**; the 29 unverifiable are all in the five reverted drafts, which have no anchors again, and the 2 ungrounded are the extraction artifacts described below. Coherence: three passes, no in-scope contradictions.

Three caveats I would rather state than imply:

- **The coherence check samples rather than exhausts.** Three passes over these drafts returned 2, 3 and 2 pairs with *different membership* each time. A robust contradiction recurs; a subtle one may surface once in three. So it is run repeatedly here, and a single clean pass should not be read as proof — by you or by me.
- **Its mechanical gate proves the quotes are real, not that they contradict.** That judgement is still a model's, and a human has to confirm it.
- **Two residual `ground-claims` flags are extraction artifacts, not defects**, and I would rather show the working than quietly filter them: `sms.clear_failing_schedules` in 10442 is reached via `config.get('sms')?.clear_failing_schedules`, and `smsparser.parse` in 10729 is how *callers* reference an export that is spelled `exports.parse` inside the file.

agent-memory/domains/messaging/issues/9022-fix6383-use-correct-sms-context-in-webapp-and-sentinel.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Root Cause
+
+Both paths already attached `patient: doc.patient` and `place: doc.place` — but only when a shortcode id was present on the doc's `fields`. In shared-libs/transitions/src/schedule/due_tasks.js, `getTemplateContext` short-circuited with `return Promise.resolve();` (an **undefined** context) whenever neither `doc.fields.patient_id` nor `doc.fields.place_id` was set, even though the doc had already been hydrated with `doc.patient`/`doc.place`. In webapp/src/ts/services/format-data-record.service.ts, `context.patient` was gated on `if (patientId)` where `patientId = doc.patient_id || doc.fields?.patient_id`, so a hydrated `doc.patient` whose shortcode lived only at `doc.patient.patient_id` never reached the context. The fix moves `patient`/`place` out of the conditional (Sentinel), re-gates on `if (doc.patient)` / `if (doc.place)` (webapp), and falls back to `doc.patient?.patient_id` / `doc.place?.place_id` when looking up registrations.
```
sugat009, line 52: nitpick (non-blocking, approving regardless): this opening clause says the gate was `doc.fields` only, but line 48 above states it correctly as "at the top level or under `fields`", and the rest of this same paragraph then gives `patientId = doc.patient_id || doc.fields?.patient_id`. So the clause is narrower than both its own neighbours. Dropping "on the doc's `fields`" or matching line 48's wording removes the inconsistency.

agent-memory/domains/messaging/issues/9467-rapidpro-error-handling.md
```diff
@@ -21,33 +25,37 @@ When RapidPro SMS gateway responded with a 400 error for invalid phone numbers (
...
-2. Parse the error response structure (which contains `urns` validation errors)
-3. Mark the message as `failed` instead of leaving it in a retryable state
-4. Log the validation error clearly for debugging
-5. Prevent further retry attempts for messages with invalid recipient phone numbers
+1. In `sendMessage`'s existing `.catch`, test exactly `err?.statusCode === 400` (the added comment cites https://rapidpro.io/api/v2/ — "Do not retry with the same values"). `statusCode` is the property shape at the time of this fix; the same branch on current master tests `status` instead (api/src/services/rapidpro.js:109), and the Testing bullet's `.rejects({ statusCode: 400 })` stub is of the same vintage
+2. Return `getStateUpdate(STATUS_MAP.failed, message.id, undefined)` so the message becomes `state: 'failed'`, `details: 'Failed'`, with no gatewayRef — the error body (the `urns` validation detail) is never parsed, only logged verbatim by the pre-existing `logger.error` call ("Error thrown when trying to send message: %o", rapidpro.js:110), which is untouched — the `// ignore error, sending the message will be retried later` comment moved below it, leaving the log ahead of the new branch
```
sugat009, line 46: nitpick (non-blocking): `rapidpro.js:110` here has no scope qualifier, one bullet after bullet 1 explicitly scopes `:109` to current master. Two adjacent line numbers for the same region, one scoped and one not, invites a reader to mix up which tree each refers to. Worth stating which tree `:110` is in.

PR #122, opened 2026-06-24
chore(memory): promote strong-fit forms-and-reports drafts from memory-pipeline for review

Hareet:
Promotes **47 strong-fit `forms-and-reports` drafts** from `agent-memory/_pending/` into `agent-memory/domains/forms-and-reports/issues/` for squad content review.

**Categories:** feature (21), bug (20), improvement (6)
**Themes:** Enketo widgets & XPath extensions, cht-datasource report create/update, SMS parser, form validation/submission hardening.

All 47 carry `domainFit: strong` + a `## Domain Rationale` section. 5 weak-fit drafts deferred for later.

- Base: `seeding-claude-cli-v2` (#119) — retarget to `main` after the schema lands.
- `validate-schema`: passing, 0 failures.
- Review for **content** (domain fit + distillation accuracy), not code.

PR #123, opened 2026-06-24
chore(memory): promote strong-fit tasks-and-targets drafts from memory-pipeline for review

Hareet:
Promotes **35 strong-fit `tasks-and-targets` drafts** from `agent-memory/_pending/` into `agent-memory/domains/tasks-and-targets/issues/` for squad content review.

**Categories:** feature (21), bug (9), improvement (5)
**Themes:** rules engine (banish nools, stale-state migration), aggregate targets & filters, task filtering/telemetry, reporting periods.

All 35 carry `domainFit: strong` + a `## Domain Rationale` section. 2 weak-fit drafts deferred to Stream C.


sugat009:
Heads up, this one currently has merge conflicts with its base and will need a rebase before it can merge.

agent-memory/domains/tasks-and-targets/issues/9232-feat9231-enable-place-filter-aggregate-targets-for-users-wit.md
```diff
@@ -0,0 +1,105 @@
+---
+id: cht-core-137
+category: feature
+domain: tasks-and-targets
+domainFit: strong
+issueNumber: 137
```
sugat009: **issue:** `id`/`issueNumber`/`issueUrl` = 137. That is `medic/care-teams#137` from the PR body, mis-stamped as cht-core; cht-core#137 is unrelated ("Restyle data records"). PR #9232 closes **cht-core#9231**.

agent-memory/domains/tasks-and-targets/issues/9570-fix9552-migrate-stale-target-state-for-interval-turnover.md
```diff
@@ -0,0 +1,81 @@
+---
+id: cht-core-9570
+category: bug
+domain: tasks-and-targets
+domainFit: strong
+issueNumber: 9570
```
sugat009: **issue:** 9553/9555/9569/9570 are four PRs (base + backports) for one issue **#9552**, distilled as four separate memories; they should collapse to one with `source_prs[]`.

agent-memory/domains/tasks-and-targets/issues/10390-feat10343-add-target-intervals-to-cht-datasource.md
```diff
@@ -0,0 +1,94 @@
+---
+id: cht-core-10343
+category: feature
+domain: tasks-and-targets
```
sugat009, line 4: **issue (domain-fit):** all files are `shared-libs/cht-datasource/*` plus one API controller; this is data-access layer work, tied to tasks only because the docs fetched are targets.

sugat009 (changes requested):
Content review. Prose fidelity is high (8772 reproduces the real 283s->63s benchmark exactly). Blockers:

- **issue (blocking): identity keys record the PR number on 15 of 35 drafts** (verified); `8838` closes no tracked issue.
- **issue (blocking): hallucinated identity on `9232`.** `id`/`issueNumber`/`issueUrl` = **137**, but 137 is `medic/care-teams#137` (a cross-repo link in the PR body) mis-stamped as cht-core; cht-core#137 is the unrelated "Restyle data records". PR #9232 actually closes **#9231**.
- **issue (blocking): 3 duplicate clusters (8 files):** issue 9231 = [9232, 9282, 9317], 9431 = [9486, 9549], 9552 = [9553, 9555, 9569, 9570].
- **issue: `related_issues: []` empty on every draft; `domainFit: strong` on every draft.** Forced picks: `10390`/`10432` (cht-datasource data-layer, tied to tasks only via the target docs), `8932` (touches `tasks.component.ts` and `contacts.component.ts` equally), `10786` (telemetry + `replications.js`/`purger.js` pipeline).
- **nitpick: process-narrative leaks into prose** on ~11-18 of 35 drafts (named reviewers, "CI green, all 47 checks pass", "the human reverted an AI change").

Clean bill: no secrets, no PII, schema 100% valid.

agent-memory/domains/tasks-and-targets/issues/9553-fix9552-handle-rules-engine-stale-state-after-upgrade.md
```diff
@@ -0,0 +1,111 @@
...
+- shared-libs/rules-engine/test/provider-wireup.spec.js
+- tests/e2e/default/targets/target-accuracy.wdio-spec.js
+
+## Testing
+
+Added/updated unit tests in rules-state-store.spec.js and target-state.spec.js covering the stale-state-after-upgrade case (missing target aggregate), plus an e2e WebdriverIO spec (target-accuracy.wdio-spec.js) validating that targets compute accurately after the configuration/upgrade scenario. The interval-turnover facet added/updated unit coverage in target-state.spec.js, rules-state-store.spec.js, and provider-wireup.spec.js exercising turnover detection and migration of stale target state into the current interval (PR #9569, #9570).
+
+## Backports
+
+The stale-state-after-upgrade fix was backported to the 4.1.x line (PR #9555, cherry-pick of dc47c51) and to 4.13.x. The interval-turnover migration was also backported to 4.13.x (PR #9570).
```
sugat009: **issue (accuracy):** this says the fix was "backported to the 4.1.x line (PR #9555)", but PR #9555 is the **4.13.x** backport (title "…for 4.13.x", base `4.13.x`), and there is no 4.1.x backport in `source_prs`. Looks like a merge-synthesis artifact — please correct 4.1.x → 4.13.x.

agent-memory/domains/tasks-and-targets/issues/10390-feat10343-add-target-intervals-to-cht-datasource.md
sugat009, line 4: **issue (domain):** the anchor PR extends cht-datasource (target-intervals *library* API); per the agreed `data-access` domain this belongs there with `secondaryDomains: [tasks-and-targets]`. See the detailed `data-access` proposal comment on #122.

Hareet: I think re-linking the cht-datasource tickets happens after we create the data-access domain, or included in that PR.
I'm imagining we re-link tickets and modify our data-labeling as the agent gets used, like say a domain gets added or combined. How do you see that process? 

I'm okay with either, and include the new domain and re-link inside the PRs as well.

sugat009: I think adding a new domain now would be easier than adding or modifying later on once we have a bit more of these context files.

Hareet:
Both round-2 items are fixed, and putting this branch through the same verification the configuration and messaging batches got turned up a class none of us had seen — one epic reshaped five of these drafts before it landed. Details below, worst first.

### Your two items

**9553's invented backport version** — fixed, and every number in the replacement checks out:

```sh
git -C $CORE log -1 --format='%h %s' c8a7f13
# c8a7f13 fix(#9552): handle rules-engine stale state after upgrade  for 4.13.x (#9555)
git -C $CORE branch -r --contains c8a7f13 | grep -E 'origin/[0-9.]+x$'   # origin/4.13.x only
curl -s https://api.github.com/repos/medic/cht-core/pulls/9555 | jq -r '.base.ref'   # 4.13.x
```

The draft now also records why there is no 4.1.x backport: that line's tip is `62aadbd` (4.1.2, June 2023), which predates the October 2024 fix.

**10324 and 9650 Domain Rationale scaffolding** — stripped.

### The thing worth your attention: one epic, five drafts

`10324`, `10371`, `10390`, `10436` and `10507` all merged into the `10140_previous-month-targets` feature branch and reached master only in that epic's squash, #10423 (`622c625427`). The epic reshaped them on the way, so their file paths are accurate for their own PRs and wrong as directions to anyone reading master:

| the draft names | master has |
|---|---|
| `tests/e2e/default/analytics/analytics.wdio-spec.js` | `tests/e2e/default/targets/analytics.wdio-spec.js` — the directory was renamed |
| `webapp/src/ts/libs/config.ts` (created by 10507, renamed by 10436) | nothing; only `.mocharc.js` survived, and the subtitle logic moved into `rules-engine.service.ts` |
| 10390's ten `target-interval.*` files (of the 21 it touched) | `target.*` — same feature, renamed wholesale |

```sh
git -C $CORE ls-tree --name-only -r origin/master shared-libs/cht-datasource/src | grep target
#   .../src/local/target.ts  .../src/remote/target.ts  .../src/target.ts     ← no -interval
git -C $CORE log --all --diff-filter=A -- '*target-interval*' | wc -l        # 0 — never on any branch
```

The four that stay here are now time-scoped with their landed locations. **10390 is not** — its vocabulary needs a rewrite, not a patch, and I would rather flag that than half-fix it. It does now carry the same "paths are as of this PR" banner as the others, so nobody reads those ten names as directions to master in the meantime. See below.

### Everything else the pass found

**Twelve self-contradictions across nine drafts**, the same class you identified on #120: the grounding pass corrected the sections that assert mechanism against code and left `summary` and Design Choices asserting what it had just disproved. Each verified against cht-core before rewriting — 9232 claimed the filter shows "only for multi-facility users" while its own Solution says there is no facility-count gate; 9705's Design Choices claimed the fix throws on *write* errors when it changed a read catch-all and the file's one `bulkDocs` still only `console.error`s; 10623 said filters were reused when `overdue-filter` and `task-type-filter` are new components on master.

Then **five more** that my own first round of fixes created — correcting one section and leaving a sibling saying the old thing, which is the failure mode this whole exercise is about, committed by the person fixing it. 9232 and 9553 each needed three passes before every section agreed.

Then **eighteen more across thirteen drafts** in a final round of repeated passes, each checked against cht-core rather than reasoned about. Worth naming, because two were mine and several are classes I had not been testing for:

- **10480's `summary`** said the PR generalised the unread-count *flow* to carry either count. It does not: `setBubbleCounter` still only carries reports and messages, and `getBubbleCounter` spreads that and *adds* `task: taskState.overdue?.length || 0` from a different slice. What was reused is the bubble at the display layer, not the dispatch. My own Code Patterns rewrite had said so plainly and I left the summary contradicting it.
- **10480 again**: I had asserted "a task due later today still counts" without checking why. It is true, but only because `moment(emission.dueDate, 'YYYY-MM-DD')` parses date-only, putting a task due today at midnight — already before `moment()`. The draft now states the mechanism and cites medic/cht-core#3943, which asks for exactly "Overdue" and due-"Today" and "not things that are due tomorrow".
- **10436's Testing** called two page objects new and both e2e helpers new. `git diff-tree` on its own PR says both page objects were modified, `targets-helper-functions.js` was added and `aggregates-helper-functions.js` was modified. This one `ground-claims` caught; earlier passes had simply never sampled that sentence.
- **10507** described `libs/config.ts` as "a helper this PR introduced — was updated to", both at once, and gave it a present-tense home in Code Patterns while the same draft says it is not on master.
- **10436** referred to `libs/config.ts` twice more without that scoping, and **10371's** Code Patterns implied one component records both open and selection telemetry — `analytics-filter.component.ts` records only the `:open` keys; the selection keys live in the sidebar filter component.
- **10623's** Code Patterns named reports and messages as the components that stopped duplicating lineage filtering. `tasks.component.ts` dropped its `ExtractLineageService.removeUserFacility` call too, which is the whole reason the refactor appears in a tasks PR.
- **9232's `summary`** said multi-facility users "had no way to scope the aggregate targets analytics view by place", while its own Problem section says they were blocked from viewing aggregate targets at all. medic/cht-core#9231 is explicit — "in `4.9.0`, users assigned to multiple `facility_ids` cannot view aggregate targets" — so the summary was understating a regression as a missing feature. Third time this draft's summary needed correcting.
- **10362** called the comparator `order-by-due-date-and-priority` in two sections. That is the name of its *test file*; the export is `orderByDueDateAndPriority`. The draft had both, and used the filename where the symbol belonged.
- **11142** attributed `FreetextFilterComponent` to the tasks module. It lives in `components/filters/freetext-filter/freetext-filter.component.ts` — the Tasks page reuses it, which is the actual point, so the draft now says which file it is.
- **10423's Testing** listed eight test files under one "Added". Six really were added; the two e2e WDIO specs (`analytics.wdio-spec.js`, `target-aggregates.wdio-spec.js`) already existed and were modified. Both `ground-claims` passes caught this one, which is what recurrence is supposed to look like.
- **10324's Testing** opened "Added an e2e WebdriverIO spec … and Karma unit specs …". Its PR adds no test file at all: every spec listed was modified, and the sidebar-filter spec was *renamed*. The section now says so outright.
- **9553's `summary`** blamed "a CHT upgrade that changes target configuration". The Root Cause blames the #9486 persisted-shape change, and the code agrees — `isStale` is `(state) => !state || !state.targets || !state.aggregate`, a pure shape check that never reads configuration. The trigger is upgrading across #9486, full stop.
- **10362's Domain Rationale** said the implementation "centers on the task pipeline (new task-notifications service, tasks reducer, …)" while its own Solution says the reducer is not consumed by the service and was only changed to import a comparator. Same class as everything else on this branch, found on the fourth round of passes.
- **9232** wrote "added new translation keys (api/resources/translations/messages-en.properties)". The keys are new; the file is not.
- **9553's Solution** said the migrated emissions are rewrapped "so `handleIntervalTurnover` can read them against the active interval", while Design Choices said the same function writes the *previous* interval's target doc. Design Choices is right: the function returns early when `stateCalculatedAt` falls inside the current interval, and otherwise aggregates against `calendarInterval.getInterval(monthStartDate, stateCalculatedAt)` — the stale interval — and stores that. Reading against the active interval is the one thing it never does. And true to form, my correction introduced the next finding: it named `calendarInterval.getInterval` and `stateCalculatedAt` — both real at the anchor, both gone from master because #9718 (in this batch) removed the whole mechanism. The drift check caught that in three of the next four passes; the sentence is now time-scoped and points at the 9718 draft.
- **8772's `summary`** described the short-circuit as firing "when the key count reaches 500 (the guard bypasses only `length < MAX_QUERY_KEYS`)". That is correct and almost unreadable — the parenthetical states the *inverse* condition, and two independent passes read it as contradicting the Solution. The guard is `!params?.keys || params.keys.length < MAX_QUERY_KEYS`; the summary now says so in one direction only.

I also swept all three domains mechanically for any path a draft lists that is absent from `origin/master` and then refers to elsewhere without scoping. Fifteen hits: ten are past-tense statements about what the PR did and are correct as written (10279 *removing* `emitter.nools.js` is the point of that PR), and five are present-tense claims in 3406, 4039 and 4374 — all three of them anchorless drafts outside #120's diff, already carved out. Nothing in scope on any of the three branches.

**All 25 timestamps** were stale — the relink, dedup and grounding commits never bumped them.

### 10390 is left for the data-access work

Your inline comment asks to move it to `data-access` with `secondaryDomains: [tasks-and-targets]`. I have left it here — whether it ends up moved or simply carries a secondary domain is exactly the sort of thing that taxonomy change should decide — on the strength of your own recommendation on #122:

> Ship `data-access` + `secondaryDomains` as **one coordinated schema/taxonomy PR** (alongside the agreed data-access work, not through these content chores) … **Not blocking any single PR**.

Doing it here would mean three open PRs racing to add the same enum value, and a corpus where one extender moved while the other seven did not. There is now a second reason: 10390's entire file and API vocabulary needs rewriting to the landed `target.*` names, so it wants a rewrite by someone thinking about cht-datasource, not a re-key bolted onto a tasks PR. A proposal for that work is being written up separately, and an early count suggests the datasource scatter is considerably wider than the drafts named on these three PRs. If you would rather it move now, say so and I will do it — but I think the sequencing you wrote is the right one.

### Gate

`validate-schema` 64/0. `verify-drafts --online` (identity, duplicates, near-miss symbols, leakage, cross-references, timestamps): **0 blocking, 0 warnings, 0 unverified** across all 25. `ground-claims`: **0 ungrounded, 0 unverifiable, 0 anchor-unusable, 0 unmatched code fences**, ~390 grounded per pass, all 25 anchored — the five epic children via #10423's squash `622c625427`, the other twenty at their own commits. `check-coherence`: **0 contradictions.**

Those are not single-pass numbers. On this exact content the gate ran **four `ground-claims` passes and three coherence passes**, all clean, after a sequence in which almost every pass had turned up something the one before it missed. That is the point of the count, not decoration.

**Two `stale-as-written` items remain, and they are disclosed rather than fixed** — 9232 names `analytics-target-aggregates-sidebar-filter.component.html` and its spec, true at its anchor and renamed to `analytics-sidebar-filter.*` when the 10140 epic generalised the sidebar past target-aggregates. They are what `stale-as-written` exists to record: accurate when written, moved since — and the draft's own epic banner already tells the reader its paths are PR-era.

I also re-ran configuration and messaging under the corrected tooling to be sure none of the five tool fixes weakened an earlier verdict: **configuration 99 grounded / 0 ungrounded, messaging 225 grounded / 0 ungrounded** (its 27 unverifiable are the five anchorless drafts carved out of #120, unchanged). One of the fixes actually removed a false finding that had appeared on #120's branch — 10230's `nepal-doit-sms`, where extraction asked whether `nepal-doit-sms.js` contains its own filename.

Four caveats I would rather state than have you find. The first is the one that matters.

- **Both checkers sample rather than exhaust, and I can now put a number on it.** Two `ground-claims` passes over byte-identical drafts each surfaced *one* ungrounded claim — and it was a **different claim each time** (11142's `FreetextFilterComponent`, then 10362's comparator name). Either pass alone would have reported "1 finding, fixed, done" and shipped the other. Coherence behaves the same way: 9232's and 9553's contradictions appeared in every pass, while 9022's surfaced in one of three. So the number below is from repeated passes, and **a single clean pass is weak evidence — including mine.**
- **Five sources of that noise turned out to be tool bugs, and are fixed on the tooling branch.** `check-coherence` was counting pairs whose own rationale read "These do not conflict". `ground-claims` reported a Root Cause claim as ungrounded when extraction had pinned it to a file the PR *created* — a pairing no commit can satisfy. It resolved a bare filename onto its real path but not a *path tail*, so 11142's true "`reducers/global.ts` only gains a `search?: string` field" came back as a misattribution. It accepted a multi-word noun phrase ("interval turnover") as a symbol. And it asked whether `nepal-doit-sms.js` contains its own filename. All five now resolve or discard rather than report, with tests. Each only ever turned a false alarm into silence, so no earlier verdict on any branch is weakened by them — but they do not make either checker exhaustive.
- **The coherence gate proves the quotes are real, not that they contradict.** That judgement is still a model's, and on this branch two of its findings (10371, 8772) turned out to be correct-but-unreadable prose rather than contradictions. I rewrote the prose anyway — if a checker misreads a sentence twice, a person will too.
- **Claim extraction is exhaustive only over code-shaped claims, and one useful class sits outside it.** Every repo path and backticked identifier is enumerated deterministically, so that tier is reproducible run to run; semantic claims ("the fix prevents duplicate sends") are still sampled and cannot be enumerated. Specifically: *added-versus-modified* is checked only when the model happens to sample the sentence — five of the findings above are that class. Making it deterministic means scoping "Added X, Y and Z" across a coordinated list, which I tried and abandoned at 126 false positives out of 164, so I raised the pass count instead of shipping a noisy check. Measured on 13 unchanged drafts, two extraction passes shared only 29% of their claims — which is why the mechanical tier was worth making deterministic, and why I am not claiming the semantic one is complete.

agent-memory/domains/tasks-and-targets/issues/10436-feat10318-display-previous-months-targets-when-selected.md
```diff
@@ -0,0 +1,104 @@
...
+- webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.html
+- webapp/src/ts/services/rules-engine.service.ts
+- webapp/src/ts/services/target-aggregates.service.ts
+- webapp/src/ts/libs/config.ts
+- webapp/src/ts/reducers/global.ts
+- webapp/src/css/inbox.less
+
+## Testing
+
+Broad test coverage added/updated. Karma unit specs updated for analytics-targets, analytics-target-aggregates, analytics-sidebar-filter, analytics-filter, rules-engine.service, and target-aggregates.service. A new Mocha unit-test harness was added for the webapp (webapp/tests/mocha/.mocharc.js, tsconfig.mocha.json, tsconfig.spec.json) specifically to unit-test webapp/src/ts/libs/config.ts (config.spec.ts). Of that harness only `.mocharc.js` survives on master: the two tsconfigs and `libs/config.ts` itself were removed or folded elsewhere before the epic squashed. WDIO e2e specs were updated for targets analytics and target-aggregates, updating the analytics and target-aggregates page objects, adding one helper (targets-helper-functions.js) and extending another (aggregates-helper-functions.js), plus updated e2e target-aggregates config.
```
sugat009: issue (blocking): this says the mocha harness "was added ... specifically to unit-test webapp/src/ts/libs/config.ts (config.spec.ts)", but PR 10436 removed all of it. From its own file list: libs/config.ts removed (+0/-26), tests/mocha/tsconfig.mocha.json removed (+0/-9), tests/mocha/unit/libs/config.spec.ts removed (+0/-103), and .mocharc.js modified +0/-2, deletions only. The harness pre-dated this PR; there was no adding step.

Two claims in the same sentence also fail: `webapp/tsconfig.spec.json` IS on master (it is the Angular/karma spec tsconfig at the webapp root, which this PR only added `sinon-chai` types to), and `libs/config.ts` was not "folded elsewhere" (`git grep getValueFromFunction origin/master` is zero-hit, so it was deleted outright).

agent-memory/domains/tasks-and-targets/issues/10436-feat10318-display-previous-months-targets-when-selected.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Root Cause
+
+The analytics targets / target-aggregates components and the shared analytics filter components did not fully propagate the selected reportingPeriod through to the UI subtitle/title rendering and the rules-engine target fetch. Subtitle text was effectively static (month name) rather than derived dynamically, and the filter UI lacked a selected-count and a contextual mobile back affordance.
+
+## Solution
+
+Threaded the selected reporting period (current/previous) through analytics-targets and analytics-target-aggregates components down to the rules-engine target fetch; derived the monthly target subtitle dynamically in webapp/src/ts/libs/config.ts (the PR-era location; the epic folded that helper into rules-engine.service.ts, so it is not on master) so it works for both current and previous periods (instead of showing the previous month's name on the aggregates page); added a back button in the mobile title bar when the previous month is selected; and added a selected-filter count badge to both the targets and target-aggregates pages via the shared analytics filter / sidebar-filter components, backed by global state in reducers/global.ts.
```
sugat009: issue: same inversion as line 94. This says the subtitle was "derived dynamically in webapp/src/ts/libs/config.ts (the PR-era location; the epic folded that helper into rules-engine.service.ts...)". The PR deleted that file, and `getValueFromFunction` exists nowhere on master, so there is no fold destination. Suggest describing where subtitle derivation lives on master instead.

agent-memory/domains/tasks-and-targets/issues/9553-fix9552-handle-rules-engine-stale-state-after-upgrade.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Root Cause
+
+#9486 changed the persisted `targetState` shape from a bare `{ [targetId]: ... }` map to `{ targets: {...}, aggregate: {...} }`. `rules-state-store.load` rehydrated a blob written by a pre-#9486 build without checking that shape, so the missing top-level `aggregate` key was carried forward and target values came out inaccurate or absent after the upgrade.
+
+For the interval-turnover facet, the persisted state stored target emissions without detecting that the reporting interval had changed between the cached state and the current computation time; on load after a turnover, the stale per-interval emissions were rehydrated and reused as-is rather than being re-scoped/reset for the now-current interval, so aggregation spanned the wrong period (PR #9569, #9570).
```
sugat009: issue (blocking): "the persisted state stored target emissions without detecting that the reporting interval had changed" is not the mechanism. Interval detection already existed before this work: `handleIntervalTurnover` in shared-libs/rules-engine/src/provider-wireup.js checks `moment(stateCalculatedAt).isBetween(currentInterval.start, currentInterval.end, ...)` and aggregates the previous interval via `calendarInterval.getInterval(...)`. Both are present at `fe795fb^`, i.e. before PR #9569. And #9569 added no interval logic at all: +1 line in rules-state-store.js, +9/-4 in target-state.js (the createState/migrateStaleState shape work), rest tests.

The real turnover facet is the pre-#9486 blob having no `targets` key, so `Object.keys(state.targets)` in `aggregateStoredTargetEmissions` threw, which is the stack trace in issue #9552.

Note this line is byte-identical to the pre-rework version (`git diff 1818c23 e1d0df4` does not touch it), so the surrounding rewrites left it behind.

agent-memory/domains/tasks-and-targets/issues/9553-fix9552-handle-rules-engine-stale-state-after-upgrade.md
```diff
@@ -0,0 +1,111 @@
...
+  - in-place persisted-state migration
+related_issues: []
+stale: false
+---
+
+## Problem
+
+Following an upgrade past #9486 (issue #9552), the rules engine carried forward a persisted target state written in the older schema, in which the top-level `aggregate` key does not exist. The post-upgrade code rehydrated that blob as-is, so affected installs showed inaccurate or missing target values until the state was rebuilt.
+
+A distinct facet of the same issue surfaces on reporting-interval turnover: when a target's reporting interval rolls over (e.g. the start of a new calendar month under the configured `monthStartDate`), the rules engine reused stale target state persisted from the previous interval, aggregating emissions from the prior period instead of resetting for the new interval and showing incorrect counts/percentages until the state was rebuilt (PR #9569, #9570).
```
sugat009: issue (blocking): the symptom is wrong for the same reason as line 66. This says the engine "reused stale target state ... aggregating emissions from the prior period instead of resetting for the new interval and showing incorrect counts/percentages". Issue #9552's actual report is a thrown error: `TypeError: Cannot convert undefined or null to object at Function.keys ... at Object.aggregateStoredTargetEmissions`. A crash, not silently wrong numbers. Also unchanged since 1818c23.

agent-memory/domains/tasks-and-targets/issues/9553-fix9552-handle-rules-engine-stale-state-after-upgrade.md
```diff
@@ -0,0 +1,111 @@
+---
+id: cht-core-9552
+category: bug
+domain: tasks-and-targets
+domainFit: strong
+issueNumber: 9552
+issueUrl: https://github.com/medic/cht-core/issues/9552
+title: Reconcile rules-engine persisted target state after an upgrade so newly configured target aggregates are not dropped
```
sugat009: issue: the title's "so newly configured target aggregates are not dropped" framing is not what either facet was about; both are the pre-#9486 persisted-shape problem. Something like "Detect and rebuild rules-engine persisted target state written in the pre-#9486 shape after an upgrade" matches the body. Also unchanged since 1818c23.

agent-memory/domains/tasks-and-targets/issues/9553-fix9552-handle-rules-engine-stale-state-after-upgrade.md
```diff
@@ -0,0 +1,111 @@
...
+
+A follow-up added interval-turnover handling. It is not triggered by comparing intervals: on every hydration `rules-state-store.load` runs the same shape check, and state failing it is migrated in place — emissions are preserved, not cleared, and simply rewrapped so `handleIntervalTurnover` can still read them. That function returns early when the state was last calculated inside the current interval; otherwise it aggregates against `calendarInterval.getInterval(monthStartDate, stateCalculatedAt)` — the interval the state belongs to, which is the one it then writes out — while staying compatible with documents written by older versions (PR #9569, #9570).
```
sugat009: nitpick: "after the configuration change" does not match the e2e case, which is titled "should handle old format of the rules-state-store" and sets up a pre-#9486 blob rather than a config change. Same wording appears again at line 99.

agent-memory/domains/tasks-and-targets/issues/9553-fix9552-handle-rules-engine-stale-state-after-upgrade.md
```diff
@@ -0,0 +1,111 @@
...
+- shared-libs/rules-engine/test/rules-state-store.spec.js
+- shared-libs/rules-engine/test/target-state.spec.js
+- shared-libs/rules-engine/test/provider-wireup.spec.js
+- tests/e2e/default/targets/target-accuracy.wdio-spec.js
+
+## Testing
+
+Added/updated unit tests in rules-state-store.spec.js and target-state.spec.js covering the stale-state-after-upgrade case (missing target aggregate), plus additions to the existing e2e WebdriverIO spec (tests/e2e/default/targets/target-accuracy.wdio-spec.js, modified not created) validating that targets compute accurately after the configuration/upgrade scenario. The interval-turnover facet added/updated unit coverage in target-state.spec.js, rules-state-store.spec.js, and provider-wireup.spec.js exercising turnover detection and migration of stale target state into the current interval (PR #9569, #9570).
+
+## Backports
```
sugat009, line 99: nitpick: this says the #9569/#9570 unit coverage exercises "turnover detection and migration". Detection was pre-existing (see line 66); what those PRs added coverage for is the shape check, `migrateStaleState` rewrapping a bare targets map, and `commitTargetDoc` receiving the previous interval's aggregate. Worth noting `handleIntervalTurnover` is zero-hit on master, removed by #9718.

agent-memory/domains/tasks-and-targets/issues/10362-feat9255-task-notificationupdated.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Code Patterns
+
+A dedicated Angular service (webapp/src/ts/services/task-notifications.service.ts) that reacts to RulesEngineService.contactsMarkedAsDirty and serialises the resulting task docs into the native Android notification store; it reads no NgRx state; reusable task ordering via the `orderByDueDateAndPriority` export of shared-libs/task-utils/src/task-utils.js for prioritizing what to surface.
```
sugat009: issue (blocking): `orderByDueDateAndPriority` is not an export of shared-libs/task-utils on master. `task-utils.js` exports only `setTaskState`; the comparator lives at `webapp/src/ts/reducers/tasks.ts:15`, and `task-notifications.service.ts:11` imports it from `@mm-reducers/tasks`. PR #10701 moved it (its body: "Moves tasks `orderBy` function from `shared-libs/task-utils` to the task reducer"). This is a present-tense claim that is false on master, so an agent would import from a path that no longer provides it.

agent-memory/domains/tasks-and-targets/issues/10362-feat9255-task-notificationupdated.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Solution
+
+Introduced a new webapp task-notifications.service.ts (class TasksNotificationService) that subscribes to RulesEngineService.contactsMarkedAsDirty (debounced 1s, maxWait 10s), fetches task docs with fetchTaskDocsForAllContacts(), and pushes the top 100 (sorted by the shared comparator) to the Android host app via globalThis.medicmobile_android.updateTaskNotificationStore(), together with a cap read from settings.tasks.max_task_notifications (default 8). It runs only when the rules engine is enabled and the user holds the can_get_task_notifications permission. The tasks reducer is not consumed by the service; it was only changed to import the sorting comparator that moved into shared-libs/task-utils. Added an `orderByDueDateAndPriority` comparator to shared-libs/task-utils/src/task-utils.js so the most urgent tasks surface first, wired the service into app.component.ts, added localized notification strings across all supported language properties files, and added a default app_settings.json entry to configure/enable the feature.
```
sugat009: issue: same root cause as line 65. "the sorting comparator that moved into shared-libs/task-utils" and "Added an `orderByDueDateAndPriority` comparator to shared-libs/task-utils/src/task-utils.js" were true at this PR's sha but were reversed by #10701. Worth stating master's location and keeping the round trip (reducer -> task-utils by this PR -> reducer by #10701) as explicit history if it is worth keeping at all.

agent-memory/domains/tasks-and-targets/issues/10324-feat10316-add-sidebar-filter-ui-for-previous-months-targets.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Root Cause
+
+Feature gap on the analytics targets page specifically: the target-aggregates page had carried a reporting-period sidebar filter (This month / Last month radios) since #9317 (analytics-target-aggregates-sidebar-filter.component.ts, 2024-08-14) and target-aggregates.service.ts already resolved its interval tag from a ReportingPeriod, but the targets view had no month-selection control and `fetchTargets()` on webapp/src/ts/services/rules-engine.service.ts took no reporting-period argument (it gained one in this work: `fetchTargets(reportingPeriod = ReportingPeriod.CURRENT)`).
+
+## Solution
+
+Added a new analytics-sidebar-filter Angular component opened by a filter icon on the green action bar, presenting two radio options (This month / Previous month). The selection drives the analytics-targets and analytics-target-aggregates components and the target-aggregates service to load the chosen month's data. Routing, the analytics-modules service, and CSS (inbox.less, targets.less) were updated, with support for both new (default) and old (can_view_old_navigation) navigation designs and RTL layouts.
+
+## Code Patterns
```
sugat009, line 66: issue (blocking): the radio labels are wrong, and the contrast drawn here does not exist. Both filters use the same two keys: `targets.this_month.subtitle` = "This month" and `targets.last_month.subtitle` = "Last month" (analytics-sidebar-filter.component.ts:42-45). "Previous month" is zero-hit in messages-en.properties; it is the enum value `ReportingPeriod.PREVIOUS`, not a label. So the sentence contrasting this filter with the aggregates filter's "This month / Last month" is comparing identical labels. A reader would grep for a UI string that does not exist.

agent-memory/domains/tasks-and-targets/issues/10324-feat10316-add-sidebar-filter-ui-for-previous-months-targets.md
```diff
@@ -0,0 +1,106 @@
+---
+id: cht-core-10316
+category: feature
+domain: tasks-and-targets
+domainFit: strong
+issueNumber: 10316
+issueUrl: https://github.com/medic/cht-core/issues/10316
+title: Add sidebar filter UI for viewing previous months' targets in analytics
+lastUpdated: '2026-08-02'
+summary: The analytics targets view only displayed the current month, with no way to review prior periods — unlike the target-aggregates view, which had carried a This month / Last month sidebar filter since #9317. Added a sidebar filter (filter icon on the green bar with This month / Previous month radio buttons) wired into the analytics targets and target-aggregates views.
```
sugat009: issue: the summary carries the same "This month / Previous month" label error as line 66. Both should read "Last month".

agent-memory/domains/tasks-and-targets/issues/10324-feat10316-add-sidebar-filter-ui-for-previous-months-targets.md
```diff
@@ -0,0 +1,106 @@
...
+
+The sidebar filter is the existing analytics target-aggregates sidebar filter generalised: analytics-target-aggregates-sidebar-filter.component.ts/html were replaced by analytics-sidebar-filter.component.ts/html (AnalyticsSidebarFilterComponent), which adds showFacilityFilter and telemetryKey inputs so the targets page can embed it with [showFacilityFilter]="false". Filter trigger is wired through analytics-filter.component.ts; month-scoped data retrieval was added to target-aggregates.service.ts and surfaced via analytics-modules.service.ts.
+
+## Design Choices
+
+Reused the established sidebar-filter pattern for UI consistency; radio buttons enforce mutually-exclusive month selection; explicitly validated for new and old navigation designs and RTL languages for backward compatibility.
+
+## Related Files
+
+> **Paths are as of this PR, not as of master.** This change merged into the `10140_previous-month-targets` feature branch and reached master only in that epic's squash, medic/cht-core#10423 (`622c625427`), which renamed and relocated several of the files below. The e2e suite moved from analytics/ to targets/.
```
sugat009: nitpick: the e2e `analytics/` -> `targets/` rename was not done by the #10423 epic squash; it happened on master in PR #10480 (commit bed454652). The master path itself is right. Suggest dropping the attribution and keeping the location. Same at line 91.

agent-memory/domains/tasks-and-targets/issues/10371-feat10332-add-targets-filter-selection-telemetry.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Solution
+
+Added selection telemetry in analytics-sidebar-filter.component.ts (collectFilterSelectionTelemetry records sidebar_filter:analytics:<telemetryKey>:<facility|reporting-period>:select), with the key supplied per page via the telemetryKey input set in analytics-targets.component.html; analytics-filter.component.ts still records only open events and gained sidebar_filter:analytics:targets:open alongside the existing target_aggregates one. E2e assertions were added on top of the pre-existing tests/utils/telemetry.js helper.
+
+## Code Patterns
+
+Instrument filter usage with a `sidebar_filter:analytics:<tab>:<action>` telemetry key, split across the two components that own each half — analytics-filter.component.ts records the `:open` events when the sidebar is opened, and the sidebar filter component records the selection events (`:facility`, `:reporting`); assert emitted telemetry entries in e2e using the shared helper at tests/utils/telemetry.js (pattern mirrored from tests/e2e/default/contacts/duplicate-contacts.wdio-spec.js).
```
sugat009: issue: Code Patterns names the telemetry segments `(:facility, :reporting)`, but line 57 of this same draft and the code use `:reporting-period`. The full key is `sidebar_filter:analytics:<telemetryKey>:reporting-period:select`. A wrong telemetry key is greppable and actionable, so worth fixing even though it contradicts only its own neighbour.

Hareet:
### Round 5 re-review request:

All five items are fixed. Each was re-derived from the PRs' own diffs and cht-core
master before editing rather than taken on your word, and all five checked out.

Then I did the sweep you asked for in the review body — *"every 'this PR added X'
and 'on master Y' sentence against `gh pr view --json files` and `git grep
origin/master`"* — mechanically, over all 25 drafts. It found one more instance of
exactly your class, and the gate found five self-contradictions on drafts you did
not name. Details worst-first.

## Your five

**10436, the mocha harness (blocking).** You were right that the PR removed all of
it; the draft said it added it. From its own file list: `config.spec.ts` +0/-103,
`tsconfig.mocha.json` +0/-9, `.mocharc.js` +0/-2, `libs/config.ts` +0/-26 — all
deletions. Also fixed the two claims in the same sentence: `webapp/tsconfig.spec.json`
is the webapp-root karma tsconfig, is on master, and only gained `sinon-chai` types
here (+2/-1); `getValueFromFunction` was deleted outright, not folded.

The replacement says where subtitle derivation actually lives, which is your
suggestion on line 64: this PR added `getReportingMonth(settings, reportingPeriod)`
to `rules-engine.service.ts` (master :588), surfaced per target as `reportingMonth`
and rendered via `subtitle_translation_key`.

**9553, the symptom and mechanism (blocking ×2).** Both were inverted, and both
were byte-identical to the pre-rework version exactly as you noted. #9552 is a
crash, not silently wrong numbers:

```sh
git -C $CORE show 'dc47c51^:shared-libs/rules-engine/src/target-state.js' | grep -n 'Object.keys(state.targets)'
#   43, 67, 227 — a pre-#9486 bare map has no .targets, so Function.keys throws
git -C $CORE show fe795fb --stat | tail -6
#   +1 rules-state-store.js, +13/-6 target-state.js, rest tests — no interval logic
git -C $CORE show 'fe795fb^:shared-libs/rules-engine/src/provider-wireup.js' | sed -n '313,332p'
#   handleIntervalTurnover, isBetween(currentInterval...), getInterval(...) all already there
```

So the turnover facet is the same shape bug: `handleIntervalTurnover` read the
persisted blob to write the previous interval's target doc and hit the same throw,
and the base fix's discard-and-rebuild would have dropped those emissions first.
Title, summary, Problem, Root Cause, Solution, Testing and Related Issues now all
tell that story, and the e2e case is named — *"should handle old format of the
rules-state-store"* — instead of being described as a configuration change. Your
line-99 note is in too: what #9569/#9570 added coverage for is the shape check,
`migrateStaleState`, and `commitTargetDoc` receiving the previous interval's
aggregate. (One extra: Code Patterns credited the removal to `#9714`, which is the
issue; the PR is `#9718`.)

**10362, the comparator (blocking).** Confirmed on master — `task-utils.js` exports
only `setTaskState`, the comparator is `reducers/tasks.ts:15`, and the service
imports it from `@mm-reducers/tasks` at `task-notifications.service.ts:11`. The
draft now states master's location and keeps the round trip (reducer → task-utils
by this PR → reducer by #10701) as explicit history, with #10701's own reason:
task-utils is for the SMS tasks attached to reports, not rules-engine tasks.

**10324, the radio labels (blocking).** Confirmed: both filters use
`targets.this_month.subtitle` / `targets.last_month.subtitle` = "This month" /
"Last month" (`analytics-sidebar-filter.component.ts:42-45`), and "Previous month"
is zero-hit in `messages-en.properties`. Summary and Solution now say "Last month"
and note that `ReportingPeriod.PREVIOUS` is an enum value, not a label.

**10371, the telemetry segment.** `:reporting-period`, matching
`collectFilterSelectionTelemetry('reporting-period')` and the full key its own
Solution already quoted.

## The sweep, and what it caught

Your closing note was the useful part of the review, so I mechanised both halves
and ran them over all 25 drafts.

*Path half* (`git grep origin/master`): every path a draft names that is absent
from master must sit on a line that scopes it. 49 hits, all accounted for —
relative fragments that do resolve, banner-scoped epic paths, the deliberately
deferred 10390, and 9232's two disclosed `stale-as-written` entries. Nothing new.

*Attribution half* (`gh pr view --json files`, via the API): every add/remove/modify
claim checked against the PR's real file status, and every "this PR added `symbol`"
claim checked against the PR's own added lines. To show the screen has teeth rather
than just returning green, I ran it against the **pre-fix** text and it reproduces
your blocking finding unaided:

```
10436:43  [webapp/src/ts/libs/config.ts]   prose says ADDED/new, PR says REMOVED
10436:43  [tsconfig.mocha.json]            prose says ADDED/new, PR says REMOVED
10436:43  [config.spec.ts]                 prose says ADDED/new, PR says REMOVED
```

On the fixed corpus the file-level screen returns 64 candidates and the symbol-level
screen 12, and every one is the "added a *symbol* to an existing file" shape —
"Added a `dbQuery` wrapper **in** pouchdb-provider.js" — or a removal correctly
attributed to a different PR. I adjudicated all 76 by hand.

It did find one real instance of your class, on a draft you had already reviewed:

**10324 credited itself with the sidebar's `telemetryKey` input.** It never touches
it. `git log`-free check, straight from the two PRs' diffs:

```sh
# 10324's additions to the component: userFacilities, showFacilityFilter — no telemetryKey
# 10371's additions: @Input() telemetryKey: string = 'target_aggregates';
```

Fixed, and it now points at #10371.

Worth flagging one thing the sweep surfaced that I deliberately did **not**
enshrine, because it is the archaeology you said not to file. #10371's diff binds
`targetModuleId = "targets"` in `analytics-targets.component.html` while the
`@Input` it adds in the same PR is `telemetryKey` — the attribute matched nothing
at that commit and was corrected to `telemetryKey` later on the feature branch, so
no commit on master ever contains `targetModuleId`. Recording the dead name would
mislead a reader; the draft states the landed form and stops there.

## Five more the gate found on drafts you did not name

**9486** took three passes because my first two fixes were each wrong in a new way
— the failure mode this whole exercise is about, committed twice by the person
fixing it. The summary said targets were computed "only ... when visiting specific
pages" while its own Root Cause described two 120s ensure-freshness debounces
(`ENSURE_FRESHNESS_SECS = 120` at `dc3ef42ab^`, used at :133 and :148). Calling
that "a 120s debounce" then contradicted "two separate"; fixing *that* left Problem
asserting nothing ran on incoming changes while Root Cause said the hook called
`updateEmissionsFor` per change. At `dc3ef42ab^` that function body is
`.then(contactIds => rulesStateStore.markDirty(contactIds))` — it invalidates, it
does not recompute. Problem now says precisely that.

**9705** — the Problem opened by putting the swallowed error on the *write* while
the summary, its own next sentence, Root Cause, Solution and Design Choices all put
it on the *read* that precedes the write. The read is right.

**10390** — Domain Rationale said "all changed files are shared-libs/cht-datasource
plus one API controller"; the PR also changes `api/src/routing.js`, adds integration
tests under top-level `tests/integration/`, and touches `.mocharc.js` and
`package.json`. (Caught twice: once as filed, once after my own first correction
still under-counted.)

**8932** — the title called the flashed strings "empty states" while Root Cause
shows both render only when the has-items flag is *true* (`hasTasks`,
`hasContacts`), which makes them end-of-list messages — the opposite. Title and two
tags follow.

**10371** — after the segment fix, the Solution read as though everything new was
on the targets page while summary and Domain Rationale said target aggregates. The
input defaults to `target_aggregates`, so the aggregates page gets selection
telemetry with no wiring and the targets page overrides the key; all three sections
now say that.

## 10390 still stays put

Unchanged from last round and for the same reasons: the taxonomy change is your own
sequencing recommendation on #122, and 10390 additionally needs its whole
`target-interval.*` vocabulary rewritten to the landed `target.*` names — a rewrite
by someone thinking about cht-datasource, not a re-key bolted onto a tasks PR. The
epic banner keeps anyone from reading those ten names as directions to master.

## Gate

`validate-schema` 89/0/3-skipped. `verify-drafts --online` **0 blocking, 0 warnings,
0 unverified** across all 25. `ground-claims` **0 ungrounded, 0 unverifiable, 0
anchor-unusable, 0 unmatched code fences**, all 25 anchored — the five epic children
via #10423's squash, the rest at their own commits. `check-coherence` **0
contradictions.**

The sequence matters more than the totals. Across the round the gate ran **14
`ground-claims` passes and 15 coherence passes**. The first of each turned up 4
ungrounded claims and 3 contradictions, and almost every pass after that found
something the one before it had missed — three separate rounds on 9486, two on
10390, and one on a draft I had already "finished" twice. It went quiet only at the
end, on frozen bytes, and the last five passes were run against the exact committed
content.

Two gate findings were the gate misreading itself, and they are worth naming
because anyone re-running this will hit them.

The first changed the prose anyway. A pass flagged
`sidebar_filter:analytics:<telemetryKey>:reporting-period:select` as a symbol that
does not exist — correctly, since it is a placeholder form no literal `git grep`
can match. It only existed because *I* had written a concrete-looking key with a
placeholder inside it. 10371 now quotes the real template literal, which is
byte-identical at the anchor and on master
(`analytics-sidebar-filter.component.ts:125` both places).

The second I left alone: one pass in fourteen flagged `target-interval` as
non-existent in 10423, from the sentence *"the document type was kept as 'target'
(its actual type) rather than 'target-interval'."* The symbol's absence is the
thing the sentence asserts, so the probe is confirming the claim while reporting it
as ungrounded — `ground-claims` has no notion of a counterfactual. I would rather
keep the rejected name on the record, since 10390's file vocabulary is still built
entirely out of it, than reword a true sentence to satisfy a probe.

Two `stale-as-written` items remain and are disclosed rather than fixed: 9232's
`analytics-target-aggregates-sidebar-filter.component.html` and its spec, true at
its anchor and renamed when the 10140 epic generalised the sidebar. That is what
`stale-as-written` is for, and the draft's epic banner already tells the reader its
paths are PR-era. A third drift item, 8932's `setTasksList(hydratedTasks)`, I did
fix rather than disclose — the local is gone from master, so the name was purely
greppable-and-wrong with nothing gained by keeping it.

PR #131, opened 2026-06-24
chore(memory): promote strong-fit authentication drafts from memory-pipeline for review

Hareet:
Promotes **39 strong-fit `authentication` drafts** from `agent-memory/_pending/` into `agent-memory/domains/authentication/issues/` for squad content review.

**Categories:** feature (15), bug (15), improvement (9)
**Themes:** SSO/OIDC login flows, token-login, permission checks & role revocation, password-reset/edit guards, multi-facility users.

All 39 carry `domainFit: strong` + a `## Domain Rationale` section. 4 weak-fit drafts deferred for later.


sugat009 (changes requested):

One precise note up front, because the intuitive version of this is not what is happening: `issueNumber` and `issueUrl` always agree with each other (the URL is literally `/issues/<issueNumber>`). The bug is that in **19 of 39** drafts here, that shared number is a **pull request**, so `/issues/N` silently redirects to `/pull/N`. The real resolved issue survives only in the PR-title slug. For example:

- `8675-feat6530-add-rate-limiting...` stores `issueNumber: 8675` / `issueUrl: .../issues/8675`, but **8675 is the merge PR** ("feat(#6530): add rate limiting for authentication requests"). The real issue is **#6530** (closed): "Add rate limiting to authentication endpoints".
- `10414-fix6784-safari-unsupported-browser...` stores `issueNumber: 10414` / `issueUrl: .../issues/10414`, but **10414 is the merge PR** ("fix(#6784): safari unsupported browser message in login page"). The real issue is **#6784** (closed): "Alert Safari users CHT doesn't support their browser".

**Duplicates in this PR.** Because each draft is keyed by its PR, some resolved issues appear more than once:
- Issue **#8868** is promoted twice (`8924`, `8933`).
- Three issues are also promoted in the contacts seeder #132 (cross-domain duplicates): **#9835**, **#9065**, **#6543**.

**Also.** One draft (`8843`) comes from a PR titled `feat(na): ...` with no linked issue at all, so there is no real issue for it to reference.

Hareet:
Ready for review. @sugat009 
Rebased onto main and addressed the review: identity keys relinked via #129's tool (19 files; all 39 PR→issue mappings re-verified against the cht-core API). The #8868 backport pair is collapsed with source_prs[]; 8843 (feat(na), closes no tracked issue) is dropped per the skip-and-flag policy. Cross-domain dupes resolved: #6543 is canonical here (the webapp display facet from the contacts seeder is folded in), while #9065/#9835/#9203 are canonically in the contacts corpus — their auth-side PR refs (#9205/#10222/#9204) are recorded there on #132. One correction from verification: 9955 was keyed correctly all along — its PR body explicitly closes #9735, the SSO epic.

agent-memory/domains/authentication/issues/8738-fix8730-fix-permission-checks-for-contact-fab-and-actionbar.md
```diff
@@ -0,0 +1,89 @@
+---
+id: cht-core-8730
+category: bug
+domain: authentication
+domainFit: strong
+issueNumber: 8730
+issueUrl: https://github.com/medic/cht-core/issues/8730
+title: Fix permission checks for contact FAB and actionbar by adding a separate con_create_people check
```
sugat009, line 8: **issue (accuracy, high):** the draft uses the permission `con_create_people` (including in a tag), but the real permission is **`can_create_people`** (PR #8738 uses `can_create_people`/`can_create_places` throughout; `grep con_create_people` = 0). Replace everywhere.

agent-memory/domains/authentication/issues/9833-feat9765-add-sso-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Root Cause
+
+Missing capability: the login controller and API had no route or service to delegate authentication to an external OIDC provider, exchange an authorization_code for an id_token, or reconcile the federated identity against an existing CHT user account.
+
+## Solution
+
+Added an OIDC SSO flow. A new openid-client.js service wraps the openid-client library to exchange the provider's authorization_code for a validated id_token. A new sso-login.js service maps the token claims to an existing CHT user and creates a session. A getOidc handler in login.js serves the GET /medic/login/oidc?code=... redirect-back endpoint (wired up in routing.js). When no matching CHT user is found, the user is redirected back to the login page with a localized helpful error message. Login UI (template, script, style) and translations across multiple languages (ar, bm, en, es, fr, hi, id, ne, sw) were updated for the SSO entry point and error messaging.
```
sugat009, line 65: **issue (accuracy):** names a `getOidc` handler in `login.js`, but the handler is **`oidcLogin`** (plus `oidcAuthorize`); `grep getOidc` = 0 in api/src. Rename.

agent-memory/domains/authentication/issues/9901-feat9764-update-tokenlogin-flow-to-not-allow-oidc-user.md
```diff
@@ -0,0 +1,88 @@
+---
+id: cht-core-9764
+category: improvement
+domain: authentication
+domainFit: strong
+issueNumber: 9764
+issueUrl: https://github.com/medic/cht-core/issues/9764
+title: Block OIDC-configured users from authenticating via the token_login flow
+lastUpdated: '2026-06-22'
+summary: Users with an oidc_provider set could still complete the token_login flow and receive a valid session, bypassing the intended OIDC-only authentication path. The token login flow now rejects such users before any session is issued.
```
sugat009, line 10: **issue (accuracy):** says the guard inspects the user's `oidc_provider` field, but the user field is **`oidc`** (a truthy flag); `oidc_provider` is the app_settings provider config (holding `client_id`), not a user property. Correct the field reference.

agent-memory/domains/authentication/issues/9900-feat9761-add-support-for-sso-users-to-admin-app.md
```diff
@@ -0,0 +1,99 @@
+---
+id: cht-core-9761
+category: feature
+domain: authentication
+domainFit: strong
+issueNumber: 9761
+issueUrl: https://github.com/medic/cht-core/issues/9761
+title: Add support for creating/editing SSO (OIDC) users in the admin app via an oidc_provider toggle
```
sugat009, line 8: **issue (accuracy):** says a "checkbox toggle sets the user's `oidc_provider`", but the PR adds a **text input** (`<input id="sso-login" type="text">`) that sets **`oidc_username`**; `oidc_provider` is only the app_settings gate. Correct the control and field.

agent-memory/domains/authentication/issues/10795-fix10341-revoke-permissions-when-user-role-is-deleted-from-a.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Root Cause
+
+All three permission-checking codepaths (shared-libs/cht-datasource/src/auth.js, shared-libs/user-management/src/roles.js, api/src/auth.js) evaluated a user's roles only against app_settings.permissions and never cross-referenced app_settings.roles. Deleting a role removed it from roles but left dangling references in permissions that were still treated as valid.
+
+## Solution
+
+Before evaluating any permission, filter the user's effective roles to only those still present in app_settings.roles. Added filterRolesByConfigured() in cht-datasource and an optional chtRolesSettings param to hasPermissions()/hasAnyPermission(); user-management hasPermission() reads config.get('roles') and filters; api hasPermission() applies the same filtering; webapp cht-datasource.service.ts now extracts and forwards roles from app settings alongside permissions. An empty or absent roles config falls back to no filtering for backwards compatibility. The duplicated logic was consolidated toward cht-datasource.
```
sugat009, line 64: **issue (accuracy):** says a `chtRolesSettings` param was added to `hasPermissions()`/`hasAnyPermission()`, but those became curried `(ctx) => (permissions, userRoles)` and read roles internally via `ctx.settings.getAll()` (`filterRolesByConfigured`). `chtRolesSettings` exists only on that internal helper, not on the public functions. Correct the API shape.

PR #133, opened 2026-06-24
feat(#114): dockerize cht-agent with git/credential sandbox

### LINKED ISSUE #114: Dockerize cht-agent + git/credential sandbox

Hareet:
Closes #114.

Containerizes the cht-agent so the pipeline runs inside an isolated sandbox on
the shared `cht-agent-net` network. The agent can read public git remotes and
edit a mounted cht-core working copy, but cannot push, reach host credentials,
or run Docker — the host operator owns all pushes and all CHT/Docker bring-up.

### Scope (maps to the issue)
- **Operator runbook** `docs/docker-agent-runbook.md` — bootstrap → build →
  verify → run → human-gated CHT bring-up → review/push → teardown, plus a
  Security Model section.
- **Image** `docker/Dockerfile` — `node:22-bookworm`, cht-conf baked in,
  multi-layer git/push/credential hardening, non-root `agent` user (`ARG AGENT_UID`),
  code-gen tooling extension point, HEALTHCHECK.
- **Compose** `docker/docker-compose.cht-agent.yml` — external `cht-agent-net`,
  RO `.credentials.json` mount, RO shadow of the working copy `.git/config`,
  `restart: "no"`, `no-new-privileges` + `cap_drop: ALL`, resource limits, no
  Docker socket.
- **Sandbox scripts** `docker/scripts/{agent-entrypoint,init-agent-git,bootstrap-workspace}.sh`
  — startup verification that refuses to start if a layer is down, host-side
  workspace bootstrap, dummy git identity, pre-push hook.
- **Agent config** `docker/agent/{CLAUDE.md,settings.json}` — sandbox rules +
  allow/deny permissions.

### Acceptance criteria
Follow the operator runbook through step 3:
- [ ] Image builds and the container comes up; `docker logs cht-agent` ends in `Sandbox verified.`
- [ ] Sandbox security probes pass (runbook §2 "Verify by hand"): public clone/fetch works;
      `git push` from both a fresh `/tmp` clone **and** the working copy hard-fails with
      `remote-error is not a git command` (no username prompt); `docker ps` and `ssh -V`
      → command not found; no socket, no credentials.
- [ ] Research sample ticket runs end-to-end using your `ANTHROPIC_API_KEY`:
      ```
      ANTHROPIC_API_KEY=sk-ant-... CHT_CORE_PATH=$HOME/src/cht-core \
        docker compose -f docker/docker-compose.cht-agent.yml up -d --build
      docker exec -it cht-agent bash -lc 'cd /app && npm run research -- tickets/simple-example.md'
      ```
      → produces research output for the ticket.
- [ ] `cht --version` resolves in the image (cht-conf available for the Test Environment Layer).

### Security model (see runbook)
The config/hook/deny layers are accident-prevention; the real push-block is
credential absence. Use a scoped, disposable model key (never personal OAuth).
Network-egress restriction for fully-untrusted runs is tracked as a follow-up (#<egress-issue>).

### Dependency
Builds independently. The end-to-end CHT-on-network path also needs
`docker/cht-agent-net.override.yml`, which ships with #66.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+# Build context for docker/Dockerfile (context is the repo root).
+# agent-memory/ is intentionally INCLUDED — it is the context database the
+# agent reads at runtime.
+.git
+.claude
+.idea
+node_modules
+dist
+coverage
+.nyc_output
+outputs
+*.log
+.env
+.DS_Store
+*.tmp
+*.bak
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,55 +8,42 @@
 
 FROM node:22-bookworm
 
-# Tools the agent drives. openssh-client is purged: all remote reads go over
-# HTTPS, and without an ssh binary git ssh:// URLs fail closed.
-RUN apt-get update && apt-get install -y --no-install-recommends \
-    git \
-    curl \
-    bash \
-    jq \
-    build-essential \
-    python3 \
-    ca-certificates \
-    && apt-get purge -y openssh-client 2>/dev/null || true \
-    && rm -rf /var/lib/apt/lists/* \
-    && ! command -v ssh   # fail the build closed if an ssh binary survived
-
-# cht-conf baked in: the Test Environment Layer (#66) runs the `cht` binary
-# against a CHT instance over HTTP — no Docker needed inside the container.
-RUN npm install -g cht-conf
-
-# --- Git sandbox layer 2: push hard-blocked at system git config ---
-# pushInsteadOf rewrites only PUSH urls, so fetch/clone of public repos stays
-# allowed. System-level config requires root to change; the agent user cannot.
-# --add is required: pushInsteadOf is multi-valued, and a plain `git config`
-# set would clobber to the last value, silently dropping the https rewrite.
+# Root-level setup in one layer: tools the agent drives, the system-level git
+# push block, and removal of all ssh client config.
+# - openssh-client is purged so git ssh:// URLs fail closed (reads use HTTPS);
+#   the trailing `! command -v ssh` fails the build closed if ssh survived.
+# - pushInsteadOf rewrites only PUSH urls, so fetch/clone of public repos stays
+#   allowed; system config needs root, which the agent user lacks. --add is
+#   required (pushInsteadOf is multi-valued; a plain set drops the https value).
 # (Adapted from the cht-core PoC .devcontainer: "no remote at all" became
 # "remote read OK, push blocked".)
-RUN git config --system --add url."error://push-blocked-by-policy".pushInsteadOf "git@github.com:" && \
-    git config --system --add url."error://push-blocked-by-policy".pushInsteadOf "https://github.com/" && \
-    git config --system --add url."error://push-blocked-by-policy".pushInsteadOf "ssh://git@github.com/" && \
-    git config --system push.default nothing && \
-    git config --system credential.helper "" && \
-    git config --system transfer.fsckObjects true
-
-# No SSH client config anywhere in the image
-RUN rm -rf /root/.ssh /etc/ssh/ssh_config.d
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        git curl bash jq build-essential python3 ca-certificates \
+    && { apt-get purge -y openssh-client 2>/dev/null || true; } \
+    && rm -rf /var/lib/apt/lists/* \
+    && ! command -v ssh \
+    && npm install -g cht-conf \
+    && git config --system --add url."error://push-blocked-by-policy".pushInsteadOf "git@github.com:" \
+    && git config --system --add url."error://push-blocked-by-policy".pushInsteadOf "https://github.com/" \
+    && git config --system --add url."error://push-blocked-by-policy".pushInsteadOf "ssh://git@github.com/" \
+    && git config --system push.default nothing \
+    && git config --system credential.helper "" \
+    && git config --system transfer.fsckObjects true \
+    && rm -rf /root/.ssh /etc/ssh/ssh_config.d
 
 # --- Git sandbox layer 1: non-root user, no credentials ---
 # AGENT_UID must match the host user that owns the cht-core working copy so the
 # bind mount is writable both ways (override with --build-arg AGENT_UID=$(id -u)).
-# The node base image owns uid 1000 — remove it before reusing the uid.
+# The node base image owns uid 1000 — remove it before reusing the uid. The
+# root-owned, read-only ~/.ssh stops the agent adding keys or ssh config.
 ARG AGENT_UID=1000
-RUN userdel -r node 2>/dev/null || true && \
-    useradd -m -s /bin/bash -u "${AGENT_UID}" agent
-
-# Root-owned, read-only ~/.ssh so the agent cannot add keys or ssh config.
-RUN mkdir -p /home/agent/.ssh && \
-    echo "# SSH disabled for the cht-agent sandbox" > /home/agent/.ssh/config && \
-    chmod 444 /home/agent/.ssh/config && \
-    chmod 555 /home/agent/.ssh && \
-    chown -R root:root /home/agent/.ssh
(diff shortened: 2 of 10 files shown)


sugat009 (approved):
## Review: dockerized cht-agent sandbox

Reviewed this in depth and also built the image and ran it to verify the claims. Approving, this is the strong, canonical version of the sandbox and the runbook is unusually honest about its own limits. Everything below is open for discussion or follow-up rather than a merge blocker, so please push back on any of it.

### Verified by building + running it

- Push block holds at both layers: `git push` from a fresh clone and from the working copy both hard-fail with `git: 'remote-error' is not a git command.` (no username prompt), exactly as the runbook says.
- No `docker`, `ssh`, or `claude` binaries in the image; no Docker socket; `openssh-client` purged with the build-closed `! command -v ssh` guard.
- The startup verifier passes all layers and ends in `Sandbox verified.`
- The one gap is real and reproduced: `curl https://example.com` from inside returns HTTP 200 and the mounted credential is readable, so outbound exfil is possible (see item 1).

### Open discussion / follow-ups

1. **Network egress + credential exfil (low for trusted use, higher for untrusted input).** Since the credential is readable inside the container and there's no egress restriction, a prompt-injected agent reading untrusted content could curl the key out. The runbook already documents this and tracks network-egress restriction as a follow-up, which feels like the right call. For internal/trusted tickets with a disposable, narrowly-scoped key this is an accepted risk; the question is mainly about sequencing the egress restriction (e.g. `internal: true` on an agent network) before pointing this at public issues/PRs. Worth confirming the threat model we're each assuming here.

2. **PR summary vs the runbook.** The top-line description says the agent "cannot reach host credentials," but the credential is mounted readable and the runbook's Security Model says so plainly. Could we align the summary to the runbook so a summary-only reader isn't given a stronger guarantee than holds?

3. **The mounted credentials file looks inert today.** The container runs the pipeline through the ChatAnthropic SDK (which auths via `ANTHROPIC_API_KEY`), and the `claude` CLI that would consume the OAuth `.credentials.json` isn't installed yet (deferred extension point). So nothing currently uses that mount, which means mounting a real OAuth token there is all risk and no benefit right now. Drop the mount until the CLI lands, or note it as a no-op? Curious whether I'm missing a consumer.

4. **settings.json allow/deny isn't enforced on the active path.** Same root as item 3: with no `claude` CLI, the allow/deny lists have no runtime effect for the SDK pipeline. Totally fine today; worth a code note so that when the CLI is added at the extension point, we re-audit the (currently agent-writable) settings file and the `--global`/`--local` git-config gap, since they become load-bearing then.

5. **Two copies of this sandbox (coordination with #119).** #119 vendored an earlier copy of this tree and they've diverged, this PR is the stronger one (adds `cap_drop: ALL`, `no-new-privileges`, the multi-valued `--add` pushInsteadOf fix, per-scheme push verification, the widened token scan). #119's own header comment already plans to delete its copy once #114 lands. Proposal: land this as the single source of truth, then point #119 at the published image and drop its vendored tree, so we don't maintain two. Does that match your intended merge order?

6. **Verifier message vs warn-only layers.** The pre-push-hook, read-only `.git/config`, and missing-working-copy checks warn rather than fail, so `Sandbox verified.` can print with one of those absent (e.g. on a UID mismatch). The system push block still holds in that case, so it's not a push enabler, but should the message be "partially verified" there, or those checks be promoted to failures?

7. **Smaller items:** the deny list covers `git config --system` but not `--global`/`--local` (and can't block the `GIT_CONFIG_SYSTEM` env bypass, the real backstop is credential absence, which is solid); the SSH-key check only globs `id_*`; and the base image is tag-pinned but not digest-pinned, so rebuilds can drift. All minor.

### What's genuinely strong

The real push-block is credential absence and it's enforced (the token/credfile scan hard-fails, backed by system+repo `pushInsteadOf` to `error://`, `push.default=nothing`, an empty credential helper, a read-only `.git/config` shadow mount, and a pre-push hook). No Docker socket, ssh purged with a build guard, non-root user, `cap_drop: ALL` + `no-new-privileges` + `restart: "no"`, selective `COPY` so host files aren't baked in, and a HEALTHCHECK. And the runbook's Security Model is refreshingly candid about exactly what is and isn't contained, which is what makes this safe to reason about. Nice work.

Approving as the canonical sandbox; happy to dig into any of the above.

Hareet:
> The mounted credentials file looks inert today.

I'm using that in further work-in-progress, so I wanted to keep that in this PR to avoid conflicts. It'll land shortly.

I'll open a further issue around 1 to unblock this merge, and everything else was addressed or can be captured in that next issue.

PR #141, opened 2026-07-11
feat(#8): wire test-generation layer into Development Supervisor

### LINKED ISSUE #8: Create Development Supervisor
Create Development Supervisor agent using LangChain under which there would be two agents:

1. Code Generation Agent
2. Test Environment Agent

Design [here](https://github.com/Hareet/cht-agent/blob/main/design/CHT%20Hierarchical%20Multi-Agent%20System%20PoC.md).

aperson30:
## Summary

Closes #8.

Integrates the Test Generation Layer into the Development Supervisor, fulfilling the core requirement of issue #8: the supervisor now orchestrates both code generation **and** test generation using LangGraph.

- Adds a `generateTests` node to the LangGraph workflow, wired after `validateImpl` and outside the refinement loop, so test generation runs exactly once on the final accepted code
- Adds `TestGenerationAgent` (wrapping a pluggable `TestGenModuleRegistry`) with a `claude-api` module that produces test files in-memory
- Adds `TestGenerationResult` type and `test-generation` phase to `DevelopmentState`
- Adds a semantic-correctness gate (planted-bug fixtures) that spawn-executes an emitted spec to prove it fails on buggy source and passes on fixed source
- Removes the legacy `TestEnvironmentAgent` stub that was standing in for this layer
- Test-gen failure is non-fatal — the workflow proceeds to HC2 even if test generation fails
- `resolveValidateImplEdge` now distinguishes warn-only plan-adherence notes from blocking issues, and ends cleanly on 0-file runs

## What changed vs the original `8-create-development-supervisor` branch

The original branch had an early, simpler version of the supervisor with the test layer stubbed out. This PR builds on top of the now-merged Code Generation Layer (#62) and brings in the full test-generation implementation, keeping the refinement loop and all hardening from main.

## Test plan

- `npm test` — 1146 passing
- `npm run build` — clean build
- `npm run lint` — clean (advisory)

## Design notes

- Test-gen output never feeds the validation score or refinement loop
- The `generateTests` node name intentionally differs from the `testGeneration` state channel (LangGraph forbids node names that collide with state channels)
- Tool use is gated on `LLMProvider.honorsCustomTools`; the CLI path disables tools to keep it capturable and prevent out-of-band writes

### DIFF
--- a/src/agents/code-generation-agent.ts
+++ b/src/agents/code-generation-agent.ts
@@ -781,7 +781,25 @@ export class CodeGenerationAgent {
    * Validate generated files
    */
   private validateGeneratedFiles(files: GeneratedFile[]): GeneratedFile[] {
-    return files.filter(file => this.normalizeGeneratedFile(file));
+    return files
+      .filter(file => this.normalizeGeneratedFile(file))
+      .filter(file => this.keepNonTestFile(file));
+  }
+
+  /**
+   * F-D: test-gen owns test authorship in all modes (matches the prompt policy).
+   * The CLI code-gen module can author `*.spec.*`/`*.test.*` via its native tools;
+   * drop those so the test-gen phase is the single author (and there is no
+   * same-path collision with test-gen's output). Anchored to the extension so it
+   * never catches `manifest.js` / `latest.js` / `contest/...` (unlike a substring
+   * check on "test"/"spec").
+   */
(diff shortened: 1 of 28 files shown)


aperson30:
Closing as a duplicate of #110, which already implements the test-generation layer. Opened in error.

PR #142, opened 2026-07-11
feat(#47): add retry logic with exponential backoff to Claude CLI provider

### LINKED ISSUE #47: Add retry logic to Claude CLI provider
## Summary

Add retry logic with exponential backoff to the Claude CLI provider for transient failures.

## Background

Network issues and rate limits can cause temporary failures. Automatic retry improves reliability.

## Requirements

- Retry on timeout errors
- Retry on rate limit errors (429)
- Maximum 3 retries (configurable via `CLAUDE_CLI_MAX_RETRIES`)
- Exponential backoff: 1s, 2s, 4s
- Log retry attempts

## Implementation

```typescript
async function invokeWithRetry(prompt: string, maxRetries = 3): Promise<string> {
  for (let attempt = 0; attempt < maxRetries; attempt++) {
    try {
      return await invoke(prompt);
    } catch (error) {
      if (isRetryable(error) && attempt < maxRetries - 1) {
        const delay = Math.pow(2, attempt) * 1000;
        console.log(`Retrying in ${delay}ms...`);
        await sleep(delay);
        continue;
      }
      throw error;
    }
  }
}
```

## Acceptance Criteria

- [ ] Retry logic implemented in `src/llm/providers/claude-cli.ts`
- [ ] Configurable via `CLAUDE_CLI_MAX_RETRIES` env variable
- [ ] Exponential backoff working
- [ ] Unit tests added
- [ ] Manual testing completed

## Dependencies

- **None** — Can start immediately

## Complexity

Medium — Error handling and async logic

aperson30:
## Summary

Closes #47.

Adds automatic retry with exponential backoff to the Claude CLI provider (`src/llm/providers/claude-cli.ts`) for transient failures, improving reliability against network blips / timeouts.

- Wraps the CLI spawn/transport step in a retry loop: **1s → 2s → 4s** backoff
- Configurable via a new `maxRetries` config option and the **`CLAUDE_CLI_MAX_RETRIES`** env var (default `3`, `0` disables)
- Logs each retry attempt (`Attempt N/M failed (...); retrying in Xms...`)
- Exposes `isRetryableCLIError()` (pure, exported) so the classification is unit-testable in isolation

## Deliberate deviation from the ticket — please note

The ticket lists "Retry on rate limit errors (429)". I intentionally **did not** retry rate-limit/auth notices, because this codebase already classifies them as **batch-fatal** (`isBatchFatalError` in `rate-limit.ts`, used by `parseResponse` + `run-pipeline`) so the batch stops *fast* rather than flagging a PR on exhausted quota. Retrying a usage-limit with a few seconds of backoff is futile (quota won't reset) and would actively undermine that documented fast-stop design.

So the retry scope is: **timeouts** (the genuine transient/network failure this provider surfaces). Not retried, by design: rate-limit/auth (batch-fatal), config errors (`ENOENT`/`EACCES`), and deterministic non-zero exits. Response parsing stays *outside* the retry loop, so batch-fatal classification is unchanged. Happy to widen the scope if the squad prefers the literal 429-retry behavior.

## Acceptance criteria

- [x] Retry logic implemented in `src/llm/providers/claude-cli.ts`
- [x] Configurable via `CLAUDE_CLI_MAX_RETRIES` env variable
- [x] Exponential backoff working (1s / 2s / 4s)
- [x] Unit tests added
- [~] "Retry on rate limit (429)" — deliberately excluded (batch-fatal by design; see above)

## Test plan

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — 1011 passing (+10 new)
- `npm run test:coverage` — all floors pass; `claude-cli.ts` at 96% statements / 90% branches

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,6 +25,8 @@ PIPELINE_CONCURRENCY=4
 # CLAUDE_CLI_TIMEOUT=300000
 # CLAUDE_CLI_MAX_TURNS=1
 # CLAUDE_CLI_SKIP_PERMISSIONS=true
+# Automatic retries (with 1s/2s/4s exponential backoff) for transient timeouts. Default: 3. Set 0 to disable.
+# CLAUDE_CLI_MAX_RETRIES=3
 
 # ── API path (only used when LLM_PROVIDER is unset) ──────────────────────
 # Anthropic API key
--- a/src/llm/providers/claude-cli.ts
+++ b/src/llm/providers/claude-cli.ts
@@ -40,7 +40,68 @@ export interface ClaudeCLIConfig {
   maxTokens?: number;
   /** Pass --dangerously-skip-permissions to the CLI. Default: true (preserves prior behavior). */
   skipPermissions?: boolean;
+  /**
+   * Maximum number of automatic retries for transient failures (default: 3, or
+   * the `CLAUDE_CLI_MAX_RETRIES` env var). 0 disables retries. Only transient
+   * errors are retried — see {@link isRetryableCLIError}.
+   */
+  maxRetries?: number;
+  /**
+   * Base delay (ms) for exponential backoff between retries (default: 1000, so
+   * 1s / 2s / 4s). Mainly a testing/tuning seam; production should leave the
+   * default.
+   */
+  retryBaseDelayMs?: number;
 }
+
+const DEFAULT_MAX_RETRIES = 3;
+const DEFAULT_RETRY_BASE_DELAY_MS = 1000;
+
+/** Resolve to a non-negative integer, falling back to `fallback` on garbage. */
+const toRetryCount = (value: number, fallback: number): number => {
+  if (!Number.isFinite(value) || value < 0) return fallback;
+  return Math.floor(value);
+};
+
+/**
+ * Resolve the retry budget from explicit config, then the
+ * `CLAUDE_CLI_MAX_RETRIES` env var, then the default.
+ */
+const resolveMaxRetries = (config: ClaudeCLIConfig): number => {
+  if (typeof config.maxRetries === 'number') {
+    return toRetryCount(config.maxRetries, DEFAULT_MAX_RETRIES);
+  }
+  const fromEnv = process.env.CLAUDE_CLI_MAX_RETRIES;
+  if (fromEnv !== undefined && fromEnv !== '') {
+    return toRetryCount(Number.parseInt(fromEnv, 10), DEFAULT_MAX_RETRIES);
+  }
+  return DEFAULT_MAX_RETRIES;
+};
+
+/** Promise-based sleep used between retry attempts. */
+const sleep = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
+
+/**
+ * Decide whether a failed CLI invocation is worth retrying.
+ *
+ * Only *timeouts* are retried: they are the transient/network failure this
+ * provider surfaces, and a fresh spawn often succeeds. Deliberately NOT retried:
+ * - Rate-limit / usage-limit and auth notices — the provider classifies these
+ *   as batch-fatal (see {@link parseResponse} / `isBatchFatalError`) so the
+ *   pipeline stops fast instead of burning backoff against an exhausted quota
+ *   or a login that won't fix itself in seconds.
+ * - Config errors (`ENOENT`/`EACCES`) — retrying a missing/unexecutable binary
+ *   is pointless.
+ * - Deterministic non-zero exits / empty results — retrying would mask, not
+ *   fix, a real failure.
+ */
+export function isRetryableCLIError(error: unknown): boolean {
+  return /timed out after/i.test(describeError(error));
+}
+
+/** Render any thrown value as a log-safe string. */
+const describeError = (error: unknown): string =>
+  error instanceof Error ? error.message : String(error);
 
 /**
  * Tools to deny when running the CLI in text-only mode.
@@ -106,6 +167,8 @@ export const createClaudeCLIProvider = (config: ClaudeCLIConfig = {}): LLMProvid
   const maxTurns = config.maxTurns ?? 20; // Multiple turns needed - test files can need 15+
   const modelName = config.model ?? 'claude-cli';
   const skipPermissions = config.skipPermissions ?? true;
+  const maxRetries = resolveMaxRetries(config);
+  const retryBaseDelayMs = config.retryBaseDelayMs ?? DEFAULT_RETRY_BASE_DELAY_MS;
   // Note: CLI doesn't support temperature/maxTokens directly via flags
   // These would be handled by account settings or model defaults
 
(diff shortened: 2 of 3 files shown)


src/llm/providers/claude-cli.ts
```diff
@@ -247,6 +310,54 @@ export const createClaudeCLIProvider = (config: ClaudeCLIConfig = {}): LLMProvid
...
+      return await executeCLI(prompt, options);
+    } catch (error) {
+      if (!canRetry(error, attempt)) throw error;
+      await backoffAfterFailure(error, attempt);
+      return null;
+    }
+  };
+
+  const executeCLIWithRetry = async (prompt: string, options?: InvokeOptions): Promise<string> => {
+    for (let attempt = 0; attempt <= maxRetries; attempt++) {
```
sugat009, line 351: issue (blocking): the loop re-spawns before the timed-out child has actually exited, which can put two tool-enabled `claude` sessions in the same working directory at once.

The timeout handler at :255-262 sends one SIGTERM, deletes the child from `activeProcesses`, and rejects, without awaiting exit or escalating to SIGKILL. The loop then sleeps only 1s and spawns the replacement. If `claude` takes longer than that to shut down (it traps signals to clean up), attempt 2 runs alongside attempt 1 in `CHT_CORE_PATH`, both with `--dangerously-skip-permissions` and tools enabled, both billing the same quota. With 3 retries that can happen 3 times per invoke.

Worse, because :258 removes it from `activeProcesses` before it is dead, the Ctrl+C handler at :183-191 can no longer reach it, so a child that ignores SIGTERM is never signalled again and outlives the parent.

**To be clear this is not your bug:** `git diff origin/main..HEAD` shows 0 changed lines touching `activeProcesses`, so that handler is pre-existing. What changes here is the consequence: a timeout used to be terminal with at most one orphan, and now it is the start of up to four overlapping sessions. That is why the fix belongs with this change.

Suggest awaiting the child's real exit before backing off (kill, wait for `close` with a grace timer, escalate to SIGKILL), and moving `activeProcesses.delete(proc)` out of the timeout callback so only the `close`/`error` handlers untrack it.

src/llm/providers/claude-cli.ts
```diff
@@ -40,8 +40,69 @@ export interface ClaudeCLIConfig {
...
+  maxRetries?: number;
+  /**
+   * Base delay (ms) for exponential backoff between retries (default: 1000, so
+   * 1s / 2s / 4s). Mainly a testing/tuning seam; production should leave the
+   * default.
+   */
+  retryBaseDelayMs?: number;
+}
+
+const DEFAULT_MAX_RETRIES = 3;
```
sugat009, line 57: issue (blocking): the default retry budget turns a hung CLI from a 10 minute failure into roughly 40 minutes, and nothing bounds the total.

`DEFAULT_MAX_RETRIES = 3` here, timeout defaults to 600000ms (:166), and each attempt re-arms the full per-attempt timeout, so worst case is 4 x 600s + 7s backoff = about 40 minutes before the error finally surfaces. On the code-gen path that lands in `development-supervisor.ts` validateImplementation, which then silently falls back to heuristic validation, so the operator waits 40 minutes for a silent degrade. With `PIPELINE_CONCURRENCY=4` a slot is blocked for that whole time.

Also note the one failure class you retry is the slowest one, so retries cost the maximum wall clock by construction.

Suggest bounding by elapsed time rather than attempt count: capture a deadline before attempt 1 and skip the retry when `now + delay + timeout` would exceed it. Failing that, defaulting `maxRetries` to 1 for a 10-minute transport, and documenting the real worst case, would both help.

src/llm/providers/claude-cli.ts
```diff
@@ -247,6 +310,54 @@ export const createClaudeCLIProvider = (config: ClaudeCLIConfig = {}): LLMProvid
...
+   * step: response parsing (and its batch-fatal rate-limit/auth classification)
+   * runs downstream in `invoke`, deliberately outside the retry loop.
+   */
+  const canRetry = (error: unknown, attempt: number): boolean =>
+    attempt < maxRetries && isRetryableCLIError(error);
+
+  // Log the transient failure and wait out the exponential backoff (1s/2s/4s
+  // at the default base) before the next attempt.
+  const backoffAfterFailure = async (error: unknown, attempt: number): Promise<void> => {
+    const delay = retryBaseDelayMs * 2 ** attempt;
```
sugat009, line 325: issue (non-blocking): the 1s/2s/4s schedule is asserted by no test. Every retry test passes `retryBaseDelayMs: 1`, so a flat delay, a zero delay, or removing the backoff entirely all keep the suite green. Since the exponential schedule is the headline of the PR title, it is worth one test that injects a fake clock or asserts the computed delays.

src/llm/providers/claude-cli.ts
```diff
@@ -40,8 +40,69 @@ export interface ClaudeCLIConfig {
...
+ *   as batch-fatal (see {@link parseResponse} / `isBatchFatalError`) so the
+ *   pipeline stops fast instead of burning backoff against an exhausted quota
+ *   or a login that won't fix itself in seconds.
+ * - Config errors (`ENOENT`/`EACCES`) — retrying a missing/unexecutable binary
+ *   is pointless.
+ * - Deterministic non-zero exits / empty results — retrying would mask, not
+ *   fix, a real failure.
+ */
+export function isRetryableCLIError(error: unknown): boolean {
+  return /timed out after/i.test(describeError(error));
```
sugat009, line 99: suggestion (non-blocking): this classifier substring-matches the whole rendered error, so a deterministic non-zero exit whose stderr happens to contain "timed out after" (a nested tool reporting its own timeout, for instance) gets retried three times. Matching on a typed error, or anchoring to the exact `Claude CLI timed out after` prefix this provider emits at :261, would be tighter.

Related: nothing fails if someone rewords that message at :261, at which point retry becomes silently dead code. One test asserting the coupling would catch it.

src/llm/providers/claude-cli.ts
sugat009, line 351: issue (non-blocking): the loop does not consult `isShutdownRequested()`, which `src/utils/shutdown.ts:7-9` documents as the contract for exactly this case: "Consumers (any module inside a retry loop) call isShutdownRequested() between iterations." Ctrl+C during a backoff window spawns a fresh session that then runs for up to the full timeout. Because both handlers use `process.once`, the SIGINT listener count is 0 afterwards, so a second Ctrl+C kills the parent and orphans that new child. Suggest checking the flag both before and after the sleep.

src/llm/providers/claude-cli.ts
```diff
@@ -40,8 +40,69 @@ export interface ClaudeCLIConfig {
...
+   * default.
+   */
+  retryBaseDelayMs?: number;
+}
+
+const DEFAULT_MAX_RETRIES = 3;
+const DEFAULT_RETRY_BASE_DELAY_MS = 1000;
+
+/** Resolve to a non-negative integer, falling back to `fallback` on garbage. */
+const toRetryCount = (value: number, fallback: number): number => {
```
sugat009, line 61: issue (non-blocking): `toRetryCount` maps every unparseable or negative value onto the default 3, so an operator who sets `CLAUDE_CLI_MAX_RETRIES=-1` or `off` meaning "disable" silently gets the full 3 retries, which is the behaviour they were trying to switch off. There is no log line revealing it. `1e3` also becomes 1 via `parseInt`. Suggest treating negative as 0 and warning on unparseable input.

src/llm/providers/claude-cli.ts
```diff
@@ -40,8 +40,69 @@ export interface ClaudeCLIConfig {
...
+  if (!Number.isFinite(value) || value < 0) return fallback;
+  return Math.floor(value);
+};
+
+/**
+ * Resolve the retry budget from explicit config, then the
+ * `CLAUDE_CLI_MAX_RETRIES` env var, then the default.
+ */
+const resolveMaxRetries = (config: ClaudeCLIConfig): number => {
```
sugat009, line 70: suggestion (non-blocking): `maxRetries` is unreachable through the repo's documented config path. `CLIProviderConfig` in `src/llm/types.ts:35-47` has no `maxRetries` field and `getCLIConfigFromEnv` in `src/llm/factory.ts` does not resolve it, so this is the only `CLAUDE_CLI_*` knob read from `process.env` inside the provider rather than through the factory. Following the existing pattern would make it settable the same way as `CLAUDE_CLI_TIMEOUT` and friends.

.env.example
```diff
@@ -25,6 +25,8 @@ PIPELINE_CONCURRENCY=4
 # CLAUDE_CLI_TIMEOUT=300000
 # CLAUDE_CLI_MAX_TURNS=1
 # CLAUDE_CLI_SKIP_PERMISSIONS=true
+# Automatic retries (with 1s/2s/4s exponential backoff) for transient timeouts. Default: 3. Set 0 to disable.
```
sugat009, line 28: issue (non-blocking): this documents only the cheap part of the cost. "1s/2s/4s exponential backoff" reads as ~7 seconds, but the real worst case is 4 x `CLAUDE_CLI_TIMEOUT` plus that backoff, so about 40 minutes at the defaults. Worth stating outright, since that is the number an operator needs.

Separately: this knob cannot be set in the one deployment that runs the CLI provider. `docker/docker-compose.seeder.yml` and `docker/Dockerfile.seeder` do not pass `CLAUDE_CLI_MAX_RETRIES` through, so "set 0 to disable" is not actually available there.

